PR #6, opened 2017-02-20
Created Boot

### LINKED ISSUE #4: Add a pxelinux.cfg config generator tool
We need a tool in metalware that generates pxelinux.cfg files based on templates in <metalwarehome>/etc/profile/ and places them in /var/lib/tftpboot/pxelinux.cfg/. 

- template looks something like
```
DEFAULT menu
PROMPT 0
MENU TITLE PXE Menu
TIMEOUT 5
TOTALTIMEOUT 5
ONTIMEOUT INSTALL

LABEL INSTALL
     KERNEL boot/centos7-kernel
        APPEND initrd=boot/centos7-initrd.img ks=http://%HOSTIP%/cluster/ks/compute.ks network ks.sendmac _ALCES_BASE_HOSTNAME=%NODE%  %KERNELAPPENDOPTIONS%
        IPAPPEND 2
```
- the template needs to be stored in /var/lib/tftpboot/pxelinux.cfg/ named after the hex value of the ip address of the node, eg ```gethostip -x node01 2>/dev/null```
- the tool can work for an individual node, or a group of nodes
- metal command would be something like ```metal boot -g groupofnodes --kernelappendoptions="HELLO=WORLD"```
- it needs to wait for a ctrl-c (much like metal hunter), when it receives ctrl-c delete all the configs thats it created.

WilliamMcCumstie:
Adds the boot option to generate pxelinux.cfg files. Closes #4


WilliamMcCumstie:
Created the template module which allows for hostip, node and kernelappendoptions
Run `metal boot --template-options` for details
Updated the boot template files to the new format

WilliamMcCumstie:
Removed kernalappend.erb, the default action is to append an empty string

Issue #8, opened 2017-02-20
kickstart template generator tool


ste78:
A kickstart template rendering tool, taking pre-defined templates from metalware tree, parsing values (for example ,<%=hostip%>) and saving to a standard location.

- pick up kickstart templates from <metalware>/etc/templates/kickstart/
- run through an ERb parser, replacing some of our already supported values
- save to /var/www/html/deployment/ks/<templatename>.ks (by default although the ability to specify the target filename (not path) would be a nice to have. 

- ```metal kickstart --template compute```  (results in <metalware>/etc/templates/kickstart/compute being read, parsed and written to /var/www/html/deployment/ks/compute.ks )

- additionally allow a set of key value pairs to be set on the command line and referenced within the template

 - example kickstart template https://raw.githubusercontent.com/alces-software/knowledgebase/master/epel/files/compute.ks

Issue #7, opened 2017-02-20
Host file tool


ste78:
We need a tool that can add entries to the host file based on templates in <metalwarehome>/etc/templates (might be sensible to start splitting up templates for each purpose - eg etc/template/boot /etc/templates/host 

- the template would look something like:
```
10.10.0.IPTAIL  NODENAME.prv.cluster.local NODENAME.prv NODENAME
10.11.0.IPTAIL  NODENAME.mgt.cluster.local NODENAME.mgt
10.11.Q3.IPTAIL NODENAME.bmc.mgt.cluster.local NODENAME.bmc
10.12.0.IPTAIL  NODENAME.ib.cluster.local  NODENAME.ib
```
- command would be something like ```metal hosts add compute --template compute --nodename node01 --iptail 101 --q3=1```

- there are many variables that could be needed in the hosts file so it might be good to start using a templating library and start passing global variables through to the template when generating

- a nice to have additional feature would be for the tool to run on node groups an 'INDEX' could be given to each itteration and maybe something like ```10.11.0.<%=@index + 100%> ``` could be used in the templates?  

- append the output to /etc/hosts (don't worry about duplicates)

PR #13, opened 2017-02-23
Feature/kickstart

### LINKED ISSUE #10: --kickstart option in metal boot
Whilst metal boot is running, render the template specified by a --kickstart option to metal boot as a temp file in /var/www/html/deployment/ks and update the ks= option in the pxeboot config files with that value. 

When the hunter process completes, as well as delete the pxeboot config file, also delete the rendered kickstart

If multiple nodes are specified with a genders group, create a kickstart file for each node and delete each one on completion.

WilliamMcCumstie:
Contains the kickstart feature that has also been integrated into boot.

Also has an updated version of Templater that supports recursive templates with arithmetic. Also fixed a file name error in hosts. Combined these two files are preventing the automerge

Fixes #10 #8 #7


WilliamMcCumstie:
Resolved conflicts

WilliamMcCumstie:
Need to fix:
Remove error when `metal kickstart` is ran
Change save file name each time `kickstart -g` is ran
Add a default file for `metal boot --kickstart`
Condense boot templates: compute.erb, login.erb, and infra.erb into install.erb
Allow the template's to be found by filename instead of full path

WilliamMcCumstie:
Issues fixed

ste78:
Can we also accept the template name without the .erb extension? ie -k compute

PR #24, opened 2017-03-03
Fixed #23

### LINKED ISSUE #23: Bug in hunter template
When trying to replace `hostip` in the hunter templates, two 6's are print as well into dhcpd.hosts.

WilliamMcCumstie:
Fixed the bug in hunter


WilliamMcCumstie:
This bug will also be fixed when `feature/yaml` is merged in. However this fix will create a conflict

Issue #17, opened 2017-02-26
Define variable hierarchy when rendering templates


ste78:
We should clarify the hierarchy of variables that are passed into templates to be rendered, this may require some refactoring of metalware to clean up the process. 
In my mind this is the basic order:

1) Constants hardcoded in metalware, eg 'hostip' is determined from a command run locally. They do not contain ERb
2) Cli user input, these come mostly from command line options, eg 'nodename'. They do not contain ERb
3) Variables defined at template parse time, eg 'index' and 'kickstartfilename'. These don't contain ERb
4) Config file input, these are whats loaded from the config file hierarchy. They may contain ERb
5) Command line json input - from the -j option for tools that support that. These may contain ERb  

Each time we parse a template, the variable hash would need to be parsed in this order, resulting in a hash of strings, which can be parsed against the template.

PR #25, opened 2017-03-07
Refactored actions and created YAML passer

### LINKED ISSUE #16: Hierarchal config file structure for ERb templating functionality
Now metalware is relying more on ERb templating for many of its tools we should look into being able to populate variables from config files, rather than have to specify everything on the cli. Ultimately this should allow us replace much of the functionality given by the puppet hiera configuration in the symphony toolset and drop puppet from deployment completely. 

I see it working something like this:

- YAML config files in <metalwarehome>/etc/config/
- config files load into a primary config hash, but in order (so individual parameters can be overwritten by config files further in the chain (much like puppet hiera) 
- hard code config load order to
1) site.yaml
2) cluster.yaml
3) %gendersgroupname%.yaml where one is available
4) %nodename%.yaml where one is available

The resulting hash would then be used when rendering any ERb template within metalware.

WilliamMcCumstie:
All actions requiring the templater have been re-factored to use a common code library. They now use the same Combiner class to ensure the hierarchy of template parameters is common and enforced. The Combiner will also search the gender file for the corresponding YAML config files.

It also contains the constant hostip which no longer needs to be provided from the actions.

Also contains test for the actions and library modules

Fixes #16 #17 #23


lib/ruby/lib/alces/stack/forkprocess.rb
```diff
@@ -0,0 +1,106 @@
+#==============================================================================
```
mjtko: This looks like it's only used in the tests too.  Another candidate for relocation?

lib/ruby/lib/alces/stack/hosts/run.rb
```diff
@@ -50,33 +51,30 @@ def run!
...
           end
           return lambda
         end
 
-        def add(json)
+        def add(template_parameters)
           append_file = "/etc/hosts"
-          json = "" if !json
-          Alces::Stack::Templater::JSON_Templater.append(@template, append_file, json, @template_parameters)
+          @json = "" if !@json
```
mjtko: A good pattern for this is: `@json ||= ""` Ref: https://github.com/bbatsov/ruby-style-guide#double-pipe-for-uninit

lib/ruby/lib/alces/stack/hunter/cli.rb
```diff
@@ -63,6 +63,11 @@ class CLI
                '--update-dhcp',
                default: false
 
+        option  :json,
+                'JSON file or string containing additional templating parameters',
```
mjtko: I can't see where this accepts a JSON file?

lib/ruby/lib/alces/stack/iterator.rb
```diff
@@ -58,11 +38,38 @@ def iterate(gender, lambda, options={})
           Nodes.new(gender).each_with_index do |nodename, index|
             options[:nodename] = nodename
             options[:index] = index
-            output << lambda.call(options)
+            output << Marshal.load(Marshal.dump(lambda.call(options)))
```
mjtko: What's this doing?

lib/ruby/lib/alces/stack/kickstart/run.rb
```diff
@@ -44,36 +45,36 @@ def initialize(template, options={})
         end
 
         def run!
+          raise "Ran from boot can only run with a single node" if @ran_from_boot and @group
           if @dry_run_flag
-            lambda = -> (json) { puts_template(json) }
+            lambda = -> (template_parameters) { puts_template(template_parameters) }
```
mjtko: I'd avoid calling the variable `lambda` as it's also a Ruby keyword.  A good alternative would be to describe what the lambda does, say, `iterator_proc`.

mjtko: Another option is to use some metaprogramming here!  You might be able to do:

```
iterator_proc = method(:puts_template)
```

lib/ruby/lib/alces/stack/templater.rb
```diff
@@ -45,99 +46,108 @@ def append(template_file, append_file, template_parameters={})
...
+          $stderr.puts "Could not parse ERB"
+          $stderr.puts template.to_s
+          $stderr.puts template_parameters.to_s
+          raise e
+        end
+      end
+
+      class Combiner < Handler
+        def initialize(json, hash={})
+          @combined_hash = {
```
mjtko: This is always going to be the same -- perhaps it could be pulled out into a constant:

```
class Combiner < Handler
  DEFAULT_PARAMS = { hostip: `hostname -i`.chomp, index: 0 }
  def initialize(json, hash = {})
    @combined_hash = DEFAULT_PARAMS.merge(hash)
    [...]
```

lib/ruby/lib/alces/stack/templater.rb
```diff
@@ -45,99 +46,108 @@ def append(template_file, append_file, template_parameters={})
...
-              print_json = true
-            elsif key.to_s == 'ITERATOR' and value.to_s == 'true'
-              print_iterator = true
+        end
+
+        attr_reader :combined_hash
+        attr_reader :parsed_hash
+
+        def load_json_hash(json)
+          return {} if !json or json.empty?
```
mjtko: Another way of writing this would be to use the tertiary operator:

```
def load_json_hash(json)
  (json || {}).empty? ? {} : JSON.parse(json, symbolize_names: true)
end
```

(Also, doesn't require the explicit `return`).

lib/ruby/lib/alces/stack/templater.rb
```diff
@@ -45,99 +46,108 @@ def append(template_file, append_file, template_parameters={})
...
+            begin 
+              yaml_payload = YAML.load(File.read("#{ENV['alces_BASE']}/etc/config/#{yaml}.yaml"))
+            rescue Errno::ENOENT # Skips missing files
+            rescue StandardError => e
+              $stderr.puts "Could not pass YAML file"
+              raise e
             else
-              none_flag = false
-              puts "    <%= #{key} %> : #{value}"
+              if yaml_payload.class != Hash
```
mjtko: A more Rubyish way of doing this is either to use `is_a?` (or `kind_of?`).

lib/ruby/lib/alces/stack/templater.rb
```diff
@@ -45,99 +46,108 @@ def append(template_file, append_file, template_parameters={})
...
-            puts "    <%= index %> : The index in the group"
-            puts "See ERB documentation for template algebra to use on 'index'"
-            puts
-          end
-          if print_json
-            puts "Include additional parameters in the JSON object. In the case of conflicts the priority order is: iterator parameters (if applicable), json inputs, then command line inputs"
-            puts
-          end
+          hash = hash.inject({}){|memo,(k,v)| memo[k.to_sym] = v; memo}
+          return hash
```
mjtko: Generally avoid explicity `return`.  This would be better as:

```
hash.inject({}) do |memo, (k,v)|
  memo[k.to_sym] = v
  memo
end
```

lib/ruby/lib/alces/stack/capture.rb
```diff
@@ -0,0 +1,53 @@
+#==============================================================================
```
WilliamMcCumstie: Moved

lib/ruby/lib/alces/stack/boot/run.rb
```diff
@@ -34,113 +34,118 @@ class Run
...
             kernelappendoptions: options[:kernel_append].chomp
           }
           @template_parameters[:nodename] = options[:nodename].chomp if options[:nodename]
           @json = options[:json]
-          @kickstart = options[:kickstart]
-          @delete_pxe = ""
+          @kickstart_template = options[:kickstart]
+          @ks_finder = Alces::Stack::Templater::Finder.new("#{ENV['alces_BASE']}/etc/templates/kickstart/")
+          @ks_finder.template = @kickstart_template if @kickstart_template
+          @to_delete = Array.new
```
WilliamMcCumstie: Fixed this instance. Will change over occurrences as I go through the code

lib/ruby/lib/alces/stack/boot/run.rb
```diff
@@ -34,113 +34,118 @@ class Run
...
           end
 
-          begin
-            Alces::Stack::Iterator.new(@group, lambda, @json)
-            kickstart_teardown if !@kickstart.to_s.empty?
-            sleep
-          rescue Exception => e
-            teardown(e)
-          rescue Interrupt
+          @e = Interrupt
```
WilliamMcCumstie: Was a throw back to how the teardown was called. Been removed

lib/ruby/lib/alces/stack/boot/run.rb
```diff
@@ -34,113 +34,118 @@ class Run
...
 
         def kickstart_teardown
           # Deletes old signal files
           delete_lambda = -> (options) { `rm -f /var/lib/metalware/cache/metalwarebooter.#{options[:nodename]}` }
           Alces::Stack::Iterator.run(@group, delete_lambda, {nodename: @template_parameters[:nodename]})
-          # 
+
+          # Gets the kick start template filename
+          ks_finder = Alces::Stack::Templater::Finder.new("#{ENV['alces_BASE']}/etc/templates/kickstart/")
+          ks_finder.template = @kickstart_template
```
WilliamMcCumstie: When I was originally designing `Finder` I wasn't sure if the same object was going to need to find multiple templates in the future. So that was major design consideration.

If that is not going to be the case for the production code (ex. tests) I'll make a change

lib/ruby/lib/alces/stack/templater.rb
WilliamMcCumstie: Updated

lib/ruby/lib/alces/stack/forkprocess.rb
WilliamMcCumstie: So I'm having issues that metalware installs a separate copy of ruby that searches the stack for the ruby files. So their are two options for moving it:

1) Create a `Alces::Stack::Test` module that contains all test specific tests. Down side is that test code is contained with the production code

2) Add a new ruby path to somewhere in the test folder (however this proved to be difficult to achieve due to how ruby is installed)

lib/ruby/lib/alces/stack/hunter/cli.rb
WilliamMcCumstie: Throw back, removed

lib/ruby/lib/alces/stack/templater.rb
WilliamMcCumstie: Fixed

lib/ruby/lib/alces/stack/templater.rb
WilliamMcCumstie: Changed

lib/ruby/lib/alces/stack/templater.rb
WilliamMcCumstie: fixed

lib/ruby/lib/alces/stack/hosts/run.rb
WilliamMcCumstie: Updated

lib/ruby/lib/alces/stack/kickstart/run.rb
WilliamMcCumstie: Fixed across all the tools

lib/ruby/lib/alces/stack/boot/run.rb
```diff
@@ -34,113 +34,118 @@ class Run
         include Alces::Tools::Execution
 
         def initialize(options={})
-          @template = Alces::Stack::Templater::Finder.new("#{ENV['alces_BASE']}/etc/templates/boot/").find(options[:template])
+          @finder = Alces::Stack::Templater::Finder.new("#{ENV['alces_BASE']}/etc/templates/boot/")
+          @finder.template = options[:template]
```
WilliamMcCumstie: Moved to initializer

lib/ruby/lib/alces/stack/boot/run.rb
WilliamMcCumstie: Template has been moved to the initializer

lib/ruby/lib/alces/stack/boot/cli.rb
```diff
@@ -73,28 +73,17 @@ class CLI
...
           options = {
             :nodename => "Value specified by --node-name",
             :kernelappendoptions => "Value specified by --kernelappendoptions",
-            :kickstart => "Determined from --kickstart (required) and nodename",
-            :JSON => true,
-            :ITERATOR => true
+            :kickstart => "Determined from --kickstart (required) and nodename"
           }
-          Alces::Stack::Templater.show_options(options)
+          Alces::Stack::Templater::Options.show(options)
```
WilliamMcCumstie: reverted

lib/ruby/lib/alces/stack/iterator.rb
WilliamMcCumstie: It creates a deep copy of the return object which could be a hash or an array depending on what is passed in. I was having issues with the return object being altered on subsequent iterations

lib/ruby/lib/alces/stack/templater.rb
```diff
@@ -169,7 +196,66 @@ def find(template)
...
+
+        class TemplateNotFound < StandardError
+          def intialize(template)
+            msg = "Could not find template file: " << template
+            super
+          end
+        end
+      end
+
+      class Options
```
WilliamMcCumstie: reverted

PR #26, opened 2017-03-08
Fixed issues in PR #25

WilliamMcCumstie:
Fixed the issues as discussed. Majority of changes are renaming variables changing Finder to except the template on initialization. 

The only issue not changed is the `Marshal.load(Marshel.dump(lambda.call(json)))` as this completes a deep copy regardless what is returned from the lambda


mjtko:
:+1:

PR #27, opened 2017-03-09
Created permanent boot feature

### LINKED ISSUE #18: Extend boot functionality to support permanent templating
Add a --permanent option to 'metal boot'

Rather than delete the pxelinux.cfg and kickstart files after the node is detected, it'll need to leave the files in place and render the pxelinux.cfg template twice (before and after detection)

There are 2 major changes to the process:

1) The kickstart files will instead of /var/lib/metalware/rendered/ks/`nodename`.ks be rendered to /var/www/html/ks/`nodename`.ks when this option is specified, when the node is detected no changes or deletions are made on this file (the kickstart file location will need to also be changed on the pxelinux template to http://<%=hostip%>/ks/

2) The pxelinux template will need to be rendered twice, a firstboot=true variable being passed to the template on the first render and a firstboot=false the second time (after the node installed routine has been triggered). After the second render, the boot process should be deemed complete and the pxelinux.cfg file left in place. The template would react something like this:
```
DEFAULT menu
PROMPT 0
MENU TITLE PXE Menu
TIMEOUT 5
TOTALTIMEOUT 5
<% if firstboot -%>
ONTIMEOUT INSTALL
<%else -%>
ONTIMEOUT LOCAL
<%end -%>

LABEL INSTALL
     KERNEL boot/centos7-kernel
        APPEND initrd=boot/centos7-initrd.img ks=http://<%= hostip %>/ks/<%= kickstart %> network ks.sendmac _ALCES_BASE_HOSTNAME=<%= nodename %> <%= kernelappendoptions %>
        IPAPPEND 2

LABEL local
     MENU LABEL (local)
     MENU DEFAULT

```
That should trigger a rebuild on the first render, and localboot thereon (but importantly allow the user to trigger a rebuild manually from the pxeboot menu system in future)

WilliamMcCumstie:
Closes #18 

Works as described in the issue


bin/metal
```diff
@@ -90,17 +90,9 @@ EOF
...
-export alces_BASE="/opt/metalware"
-export alces_CMDDIR="$alces_BASE/lib/actions"
-### RUBY
-export alces_RUBY="$alces_BASE/opt/ruby/bin/ruby"
-export alces_RUBYHOME="$alces_BASE/opt/ruby"
-export LD_LIBRARY_PATH="${alces_RUBYHOME}/lib:$LD_LIBRARY_PATH"
-export GEM_HOME="${alces_RUBYHOME}/lib/ruby/gems/2.1.0"
-export GEM_PATH="${alces_RUBYHOME}/lib/ruby/gems/2.1.0"
+export -f display_help
+$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/exportCmd
```
mjtko: I'm surprised that `bin/exportCmd` would work when executed in this way -- my expectation would be that the `export` statements in there would only affect the execution of that script, not the calling script.

I see that the motivation for this change is to allow both `metal` and `test` to share the same set of variables.  A more conventional pattern for this would be to move the variables to, say `etc/metal.rc` and then `source` that file, something like this:

```
source $(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/../etc/metal.rc
```

etc/templates/boot/install.erb
```diff
@@ -3,9 +3,17 @@ PROMPT 0
...
+<% if firstboot %>
 ONTIMEOUT INSTALL
+<% else %>
+ONTIMEOUT LOCAL
+<% end %>
 
 LABEL INSTALL
      KERNEL boot/centos7-kernel
-        APPEND initrd=boot/centos7-initrd.img ks=http://<%= hostip %>/metalware/ks/<%= kickstart %> network ks.sendmac _ALCES_BASE_HOSTNAME=<%= nodename %> <%= kernelappendoptions %>
+        APPEND initrd=boot/centos7-initrd.img ks=http://<%= hostip %><% if !permanentboot %>/metalware<% end %>/ks/<%= kickstart %> network ks.sendmac _ALCES_BASE_HOSTNAME=<%= nodename %> <%= kernelappendoptions %>
```
mjtko: Tertiary operator might be slightly neater here than the `if`:

```
<%= permanentboot ? "" : "/metalware" %>
```

WilliamMcCumstie: Updated

lib/ruby/lib/alces/stack/boot/run.rb
```diff
@@ -88,7 +91,9 @@ def save_template(parameters={})
           add_kickstart(parameters) if @kickstart_template
           combiner = Alces::Stack::Templater::Combiner.new(@json, parameters)
           save = get_save_file(combiner)
-          @to_delete << save
+          if !@permanent_boot_flag
```
mjtko: Generally better to use `unless` rather than `if !`; ref: https://github.com/bbatsov/ruby-style-guide#unless-for-negatives

Also, favour the modifier form of `if`/`unless` for single lines; ref: https://github.com/bbatsov/ruby-style-guide#if-as-a-modifier

i.e.:

```
@to_delete << save unless @permanent_boot_flag
```

lib/ruby/lib/alces/stack/boot/run.rb
```diff
@@ -114,32 +121,47 @@ def run_kickstart
...
           }
           kickstart_options[:nodename] = @template_parameters[:nodename] if @template_parameters.key?("nodename".to_sym)
           kickstart_lambda = -> (hash) {
             hash[:save_append] = hash[:nodename]
             return Alces::Stack::Kickstart::Run.new(@kickstart_template, hash).run!
           }
           kickstart_files = Alces::Stack::Iterator.run(@group, kickstart_lambda, kickstart_options)
-          @to_delete_dry_run.push(*kickstart_files) if @dry_run_flag
-          @to_delete.push(*kickstart_files) if !@dry_run_flag
+          kickstart_files
```
mjtko: This doesn't need to be here! :)

lib/ruby/lib/alces/stack/boot/run.rb
```diff
@@ -114,32 +121,47 @@ def run_kickstart
...
+          if !@permanent_boot_flag
+            @to_delete_dry_run.push(*kickstart_files) if @dry_run_flag
```
mjtko: Use `concat` rather than `push(*)`.

lib/ruby/lib/alces/stack/boot/run.rb
```diff
@@ -163,9 +182,11 @@ def teardown(e)
           tear_down_flag = false
           tear_down_flag_dry = true if @dry_run_flag and !@to_delete.empty?
           tear_down_flag = true if !@dry_run_flag and !@to_delete_dry_run.empty?
-          puts "DRY RUN: Files that would be deleted:" if !@to_delete_dry_run.empty?
-          @to_delete_dry_run.each do |file| puts "  #{file}" end
-          @to_delete.each do |file| `rm -f #{file} 2>/dev/null` end
+          if !@permanent_boot_flag
+            STDERR.puts "DRY RUN: Files that would be deleted:" if !@to_delete_dry_run.empty?
+            @to_delete_dry_run.each do |file| STDERR.puts "  #{file}" end
```
mjtko: Prefer `{...}` form for blocks when they're one-liners; ref: https://github.com/bbatsov/ruby-style-guide#single-line-blocks

etc/templates/boot/install.erb
```diff
@@ -3,9 +3,17 @@ PROMPT 0
 MENU TITLE PXE Menu
 TIMEOUT 5
 TOTALTIMEOUT 5
+<% if firstboot %>
```
mjtko: Similarly to below, this could be `ONTIMEOUT <% firstboot ? 'INSTALL' : 'LOCAL' %>`

lib/ruby/lib/alces/stack/boot/run.rb
```diff
@@ -37,7 +37,10 @@ def initialize(options={})
           @finder = Alces::Stack::Templater::Finder.new("#{ENV['alces_BASE']}/etc/templates/boot/", options[:template])
           @group = options[:group]
           @dry_run_flag = options[:dry_run_flag]
+          @permanent_boot_flag = options[:permanent_boot_flag]
```
mjtko: There are a lot of `options` now (and accompanying instance variables).  It might be worth considering extracting these into an options utility class, e.g.:

```
class Run
  def initialize(options)
    @options = Options.new(options)
  end

  class Options
    def initialize(options)
      @options = options
      # [...more as necessary...]
    end

    def template_parameters
      @template_parameters ||= {
        firstboot: true,
        permanentboot: permanent_boot_flag,
        kernelappendoptions: kernel_append.chomp
      }
    end

    def dry_run?
       !!dry_run_flag
    end

    def method_missing(s, *a, &b)
      if options.key?(s)
        options[s]
      else
        super
      end
    end
  end
end
```

With that in place you can move some of the logic into the `Options` class as method implementations and make the `Run` class a bit more readable.

WilliamMcCumstie:
```def template_parameters
      @template_parameters ||= {
        firstboot: true,
        permanentboot: permanent_boot_flag,
        kernelappendoptions: kernel_append.chomp
      }
    end
```
For this method, wouldn't the updated version of @template_parameters would be in the options scope and not in the Run scope?

mjtko: `@template_parameters` would be an instance variable contained within the instance of the `Options` class, addressable via `options.template_parameters`.  Is that what you mean?

WilliamMcCumstie:
Made those changes @mjtko

mjtko:
:+1:

PR #30, opened 2017-03-14
Created script feature

### LINKED ISSUE #19: Metalware support for installation script management
Provide basic functionality for creating customised installation scripts from templates from each node personality. 

1) 'metal script --template installsomething' should be able to take a template as input from `metalwarehome`/etc/templates/scripts/installsomething (with the same name/path abbreviating functionality as the other tools), and parse using the builtin metalware parser

2) Using a single node or group of nodes combine with the 'boot' tool by adding a '--scripts' option (ie take multiple script templates) and dump the output files to /var/lib/metalware/rendered/scripts/`node name`/`scriptname`, delete the output files when the boot complete hook is triggered. 

3) If the --permanent option is specified to 'metal boot', output is parsed and dumped to /var/www/html/scripts/`node name`/`scriptname`

WilliamMcCumstie:
Fixes #19 

Created the `metal script` action as described and linked it into `metal boot`


lib/ruby/lib/alces/stack/boot/run.rb
```diff
@@ -61,7 +62,7 @@ def ks_finder
           def template_parameters
             @template_parameters ||= {
               firstboot: true,
-              permanentboot: permanent_boot?,
+              permanent_boot: self.permanent_boot?,
```
mjtko: `self` here is superfluous... i'd just make these `permanent_boot?` calls.

lib/ruby/lib/alces/stack/boot/run.rb
```diff
@@ -100,27 +104,24 @@ def method_missing(s, *a, &b)
 
         def run!
           puts "(CTRL+C TO TERMINATE)"
-          if !@opt.template_parameters.key?("nodename".to_sym) and !@opt.group and !@opt.json 
+          if !@opt.template_parameters.key?(:nodename) and !@opt.group and !@opt.json
```
mjtko: Prefer `&&` and `||` rather than `and` and `or`. Ref: https://github.com/bbatsov/ruby-style-guide#no-and-or-or

lib/ruby/lib/alces/stack/boot/run.rb
mjtko: Oh. Ref: https://github.com/bbatsov/ruby-style-guide#no-self-unless-required 😎

lib/ruby/lib/alces/stack/scripts/run.rb
```diff
@@ -0,0 +1,84 @@
...
+        def initialize(template, options={})
+          @finder = Alces::Stack::Templater::Finder.new("#{ENV['alces_BASE']}/etc/templates/scripts/", template)
+          @group = options[:group]
+          @json = options[:json]
+          @dry_run_flag = options[:dry_run_flag]
+          @template_parameters = {}
+          @template_parameters[:nodename] = options[:nodename].chomp if options[:nodename]
+          @ran_from_boot = options[:ran_from_boot]
+          @save_location = "#{options[:save_location]}"
+          @save_location << "/" if @save_location[-1] != "/"
```
mjtko: I'd be inclined to rewrite this as:

```
@save_location = options[:save_location].chomp('/') << '/'
```

lib/ruby/lib/alces/stack/scripts/run.rb
```diff
@@ -0,0 +1,84 @@
...
+        include Alces::Tools::Logging
+        include Alces::Tools::Execution
+
+        def initialize(template, options={})
+          @finder = Alces::Stack::Templater::Finder.new("#{ENV['alces_BASE']}/etc/templates/scripts/", template)
+          @group = options[:group]
+          @json = options[:json]
+          @dry_run_flag = options[:dry_run_flag]
+          @template_parameters = {}
+          @template_parameters[:nodename] = options[:nodename].chomp if options[:nodename]
```
mjtko: Maybe squash this to the one-liner:

```
@template_parameters =
  options[:nodename] ? {nodename: options[:nodename].chomp} : {}
```

lib/ruby/lib/alces/stack/scripts/run.rb
```diff
@@ -0,0 +1,84 @@
...
+            lambda_proc = -> (template_parameters) { save_template(template_parameters) }
+          end
+
+          Alces::Stack::Iterator.run(@group, lambda_proc, @template_parameters)
+        end
+
+        def get_save_file(nodename)
+          str = "#{@save_location}".gsub("<%= nodename %>", nodename)
+          str << @finder.filename_ext_trim_erb
+          `mkdir -p #{File.dirname(str)}`
```
mjtko: This seems like a side-effect I wouldn't expect in a "getter" method.  Might be better to relocate this to where the actual saving is occurring.  I'd consider `private` methods too to distinguish between what's called externally and internally.  Here's my (untested) refactoring attempt:

```ruby
require 'fileutils'
[...]

def save_template(params)
  combiner = combiner_for(params)
  filename = filename_for(combiner.parsed_hash[:nodename])
  # Maybe even move this into the Combiner#save method?
  FileUtils.mkdir_p(File.dirname(filename))
  combiner.save(@finder.template, filename)
  filename
end

def puts_template(params)
  combiner = combiner_for(params)
  puts "SCRIPT TEMPLATE"
  puts "Hash: #{combiner.parsed_hash}"
  puts "Save: #{filename_for(combiner.parsed_hash[:nodename])}"
  puts "Template: #{combiner.file(@finder.template)}"
end

private
def filename_for(nodename)
  "#{@save_location}".gsub("<%= nodename %>", nodename) << @finder.filename_ext_trim_erb
end

def combiner_for(params)
  Templater::Combiner.new(@json, params)
end
```

lib/ruby/lib/alces/stack/templater.rb
```diff
@@ -203,14 +203,18 @@ def initialize(msg="Hash included through file method. Must be included in Combi
       class Finder
         def initialize(default_location, template)
           @default_location = default_location
-          @template = find_template(template)
-          @filename_ext = @template.scan(/\.?[\w\_\-]+\.?[\w\_\-]*\Z/)[0].chomp
-          @filename = @filename_ext.scan(/\.?[\w\_\-]+/)[0].chomp
+          @template = find_template(template).chomp
+          @filename_ext = File.basename(@template)
+          @filename_ext_trim_erb = @filename_ext.gsub(".erb", "")
```
mjtko: Maybe beter as: `File.basename(@template, '.erb')`

mjtko: Unless it doesn't end in `.erb`; not sure what this value looks like at runtime.

mjtko:
Looks good barring the few niggles above!

lib/ruby/lib/alces/stack/templater.rb
WilliamMcCumstie: `Finder` doesn't seem to have an issue with it. Ideally we would prefer being able to use non `.erb` files for the scripts because I use the file extension to determine the save file name. 

I have a few tests using `.sh` and `.sh.erb` and both seem to work fine at runtime

lib/ruby/lib/alces/stack/scripts/run.rb
WilliamMcCumstie: I was purposefully leaving it out of `Combiner`. I didn't want it to unintentionally creating files because a path is wrong. However may in the future we could create a flag that allows `Combiner` to do so

WilliamMcCumstie:
Fixed those issues

lib/ruby/lib/alces/stack/scripts/run.rb
```diff
@@ -0,0 +1,82 @@
...
+
+        def get_save_file(nodename)
+          str = "#{@save_location.gsub("<%= nodename %>", nodename)}" \
+                "#{@finder.filename_ext_trim_erb}"
+        end
+
+        def save_template(template_parameters)
+          combiner = Alces::Stack::Templater::Combiner.new(@json, template_parameters)
+          save = get_save_file(combiner.parsed_hash[:nodename])
+          FileUtils.mkdir_p(File.dirname(save))
```
mjtko: Should really have a `require 'fileutils'` at the top of this file as you're using this -- I think it's only working as one of the libraries happens to pull in `fileutils` at the moment. :)

WilliamMcCumstie:
@mjtko @ste78 Should be ready to merge

mjtko:
:+1:

PR #32, opened 2017-03-16
Release/1.1.0

### LINKED ISSUE #16: Hierarchal config file structure for ERb templating functionality

### LINKED ISSUE #18: Extend boot functionality to support permanent templating

### LINKED ISSUE #19: Metalware support for installation script management

### LINKED ISSUE #23: Bug in hunter template

### LINKED ISSUE #28: Logging
Metalware logging needs improvement so that we can better support on-site installations by being able to track what actions have been executed. Ideally into one central log file, track at least the actions being executed and major change events (files being created / deleted etc).

WilliamMcCumstie:
The next release includes:
1. Updated templater which accepts configs Fixes #16 #17 
1. Permanent booting Fixes #18 
1. Rendering of installation scripts Fixes #19 
1. Logging Fixes #28 
1. Bug Fixes #23


WilliamMcCumstie:
Need to prevent logs filling up disk space. See #33

WilliamMcCumstie:
Created rotate log file. Fixes #33

PR #37, opened 2017-03-23
Repository system

### LINKED ISSUE #29: Content Repositories
Metalware now has a large amount of configuration data / scripts / templates that can be described better as content.
This enhancement is to add functionality to refactor where this data is located into separate standalone component dirs. 

I've done an example of a 'repo' here: https://github.com/alces-software/metalware-profile-hpc

Basically this is to add a 'metalware repo' action that can 'import' a repo from a git URL into /var/lib/metalware/repos/`reponame`/. Extended functionality could also provide 'list' and 'update' functionality to this.

The idea would then be that when running commands that use templates / configs you'd use '--repo myrepo' which would use configs / templates from that repo, rather than the default one in metalware/etc/

WilliamMcCumstie:
Creates a repoing system that allows the user to set which repo to pull a template from. This can be done by using the following syntax in all the template inputs: <repo>::<template>.

The repositories are stored in `/var/lib/metalware/repos/` and can be managed directly using git or via the new `metal repo` command.

Most of the file changes are because `Finder` has been removed from the `Templater` module and placed within its own. Also `Finder` and `Combiner` has been switched from 2 to 3 inputs (each) to accommodate the repoing system.

Fixes #29


mjtko:
:+1:

PR #43, opened 2017-04-25
Updated Status command to use threads

WilliamMcCumstie:
Updated `metal status` command to use threads. This removed all the single handling and most of the kill commands. The `bash` commands still operate on their own `sub processes` however each process has it's own `Job` thread managing it in rube.

Also using the inbuilt `Queue` class in ruby and removed the old `Jobs` class. The old `Task` class has been renamed as `Job` to make the naming consistent


WilliamMcCumstie:
Also added test which makes the bulk of the additions

BobWhitelock:
Rebased this on latest `develop` for easier reviewing.

lib/ruby/lib/alces/stack/status/run.rb
```diff
@@ -19,138 +19,116 @@
...
+          while data["FINISHED"] != true
+            data = get_finished_data
+            if data.empty?
+              empty_count += 1
+              raise DataIncomplete if empty_count > 100 && @monitor.thread.stop?
+            elsif data["FINISHED"] != true
+              display_data data
+              empty_count = 0
+            end
+            sleep 0.1
```
BobWhitelock: Why the magic sleep? Could be worth adding a comment to this with the reason as the purpose isn't obvious, and similarly for other magic sleeps.

BobWhitelock:
@mjtko are you happy for this to be merged once @WilliamMcCumstie has looked at my comments, or do you want to have a look through yourself first?

WilliamMcCumstie:
Will merge the changes in with the new PR #50.

Once the current PR are merged into `feature/refactored-status` I will make the changes as described by @BobWhitelock

lib/ruby/lib/alces/stack/status/run.rb
```diff
@@ -19,138 +19,116 @@
...
+      end
+
+      def nodes
+        @nodes ||= _set_nodes
+      end
+
+      def _set_nodes
+        lambda_proc = lambda { |lambda_hash| lambda_hash[:nodename] }
+        a = Alces::Stack::Iterator.run(group, lambda_proc, nodename: nodename)
+        a.is_a?(Array) ? a : [a]
```
WilliamMcCumstie: Will need to look at the updated version of the iterator

BobWhitelock: I think this is irrelevant/resolved with the new `Nodes` class, don't worry about this

lib/ruby/lib/alces/stack/status/run.rb
```diff
@@ -19,138 +19,116 @@
...
+    end
+
     module Status
       class Run
-        include Alces::Tools::Execution
-
         def initialize(options={})
-          @opt = Options.new(options)
+          @opt = RunOptions.new(options)
+          raise "Can not specify both -g and -n" if @opt.nodename? == @opt.group?
```
WilliamMcCumstie: So the condition is correct it's the error message that is miss leading. It is intended as general input validation of the two

PR #46, opened 2017-05-22
Added options parser class

WilliamMcCumstie:
Created the `Options` parser which converts the hash input from the `cli` to a object that can be queried  form the `action`.

I believe is will be removed in the updated version of metalware but is required for `status1.2` to work


WilliamMcCumstie:
This ready to be merged in, no additional commits will be required

PR #50, opened 2017-05-22
Re-based version of `status`

WilliamMcCumstie:
Contains a rebased version of the code as seen in PR #43 

First deletes `lib/ruby/lib/alces/stack/status/jobs.rb` as this is implemented using the inbuilt ruby `queue` and is now obsolete.

`lib/ruby/lib/alces/stack/status/task.rb` has became `lib/ruby/lib/alces/stack/status/job.rb` however git is now tracking them as separate files. The name of the class was changed from `task` to `job` as to maintain consistency with the naming in the `monitor`. The other major changes are using threading instead of forking processes and reporting of completed data to a class variable.

The `monitor` has been switched over to use a thread instead of a process. It now also keeps a list of `jobs` to be completed on a ruby `queue` which removes the need of `jobs.rb` as discussed above.

The `run` process needed to be altered to use the updated version of the `monitor`. Previously the `run` process has to read completed data from a shared file. However as it now uses threading, it can read the data from the `Job` class variable as it is being wrote to.

Most of the signal handling has also been removed as it is not required for threading


WilliamMcCumstie:
Still need to make the changes as described by @BobWhitelock in #43

PR #52, opened 2017-05-23
Refactoring templater

BobWhitelock:
This PR contains a few changes to how the `Templater` interface works to support what I'm working on and to try and make the templating simpler/clearer/easier to test. See commit messages for the details of the changes.

There's quite a lot of additions/deletions here but it's mostly just churn in the tests and other files due to the interface when using the `Templater` changing; the important changes are in `metalware-new/src/templater.rb`. @WilliamMcCumstie: No need to look at this in great detail but if you want to take a quick look to see if you have any thoughts then that would be useful :slightly_smiling_face:.

### DIFF
--- a/metalware-new/spec/commands/build_spec.rb
+++ b/metalware-new/spec/commands/build_spec.rb
@@ -48,12 +48,12 @@ describe Metalware::Commands::Build do
 
   context 'when called without group argument' do
     it 'renders default templates for given node' do
-      expect(Metalware::Templater).to receive(:save).with(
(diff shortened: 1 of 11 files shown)


WilliamMcCumstie:
I haven't look at the tests in super detail so I assume they are passing or failing correctly (depending how you look at it). 

The changes in templater looks fine. Mostly just reformatting or name changes

BobWhitelock:
Great, thanks for checking.

PR #53, opened 2017-05-23
Base Command Class

WilliamMcCumstie:
I have created the base command class as discussed. Eventually this class will contain the `logging` set up and logs `fatal` errors. But this will be a latter PR 

It contains the `initialize` method that calls the `setup` function and `run` function. The base class has abstract functions for `setup` and `run` which through an error unless they are overridden. 

As a lot of the commands capture interrupt, their is a `handle_interrupt` method as well. This method by default re-raises the interrupt unless overridden.  It also will prevent interrupt from being logged (when logging is implemented).


metalware-new/src/commands/build.rb
```diff
@@ -1,6 +1,7 @@
 
 require 'active_support/core_ext/string/strip'
 
+require 'commands/basecommand'
```
BobWhitelock: The usual convention in Ruby is to name the class/module using `CamelCase` (which you've done) and then the file using `underscore_case` - very minor but can we name the file `base_command` :slightly_smiling_face:?

WilliamMcCumstie:
Hold off merging this, I just noticed the `repo` commands

metalware-new/src/commands/build.rb
```diff
@@ -113,7 +112,7 @@ def clear_up_built_node_marker_files
         FileUtils.rm_rf(files)
       end
 
-      def handle_interrupt
+      def handle_interrupt(_not_used)
```
BobWhitelock: I was a bit confused by this parameter for a moment, though I now realise it's so the signature matches that in `BaseCommand`. Could be worth giving it the same name as in `BaseCommand` to make this clear? You can still prefix it with an underscore if you prefer to indicate it's required but intentionally not used.

metalware-new/src/commands/hunter.rb
```diff
@@ -123,7 +122,7 @@ def sequenced_name
         "#{@options.prefix}#{@detection_count.to_s.rjust(@options.length, '0')}"
       end
 
-      def handle_interrupt
+      def handle_interrupt(_not_used)
```
BobWhitelock: Similarly to comment on `Build#handle_interrupt`.

WilliamMcCumstie:
Ready to for review

metalware-new/src/commands/repo/update.rb
```diff
@@ -1,13 +1,18 @@
...
 
 require 'constants'
 
 module Metalware
   module Commands
     module Repo
-      class Update
-	def initialize(_args, options)
+      class Update < BaseCommand
+	      def setup(args, options)
```
BobWhitelock: Your indentation looks messed up here.

metalware-new/src/commands/repo/update.rb
```diff
@@ -44,23 +49,23 @@ def initialize(_args, options)
             #                         "been reached!"
             raise "Internal error. Check metal log"
           end
-	end
+	      end
```
BobWhitelock: Indentation also looks shifted here.

metalware-new/src/commands/repo/update.rb
WilliamMcCumstie: Yeah the indentation was a bit screw when I opened it. Will have another pass over the files

WilliamMcCumstie:
Fixed

BobWhitelock:
Thanks!

PR #55, opened 2017-05-23
Added Logging

WilliamMcCumstie:
I have reintroduced logging into metalware.

It uses a `MetalLog` class similar to the old version however with a few updates. I was looking at the `Output` class and it maybe an idea to combine the logging and output together?

The base command by default logs the command line input (what ever is in `ARGV`) and any fatal errors.


BobWhitelock:
@WilliamMcCumstie: I've force pushed this so it uses a few fixes I've since added to `develop` for some stuff I missed previously.

Aside from what we discussed about moving the logs path to the config files/class this looks good to me. To have that work I think you should need to:

- add a `logs_path` or similar config value to `src/config.rb`;

- add a value for this within the `tmp` dir to `spec/fixtures/configs/unit-test.yaml` and `spec/fixtures/configs/integration-test.yaml`;

- ensure the dir used for testing is created before running the tests - `spec/spec_helper.rb` may be a good place to do this;

- ensure the test config is used in the different tests - I think this will be the case already for most of them, but you may need to just add a `SpecUtils.use_unit_test_config(self)` to some of the existing command specs if they don't have this already (until now when testing some commands we won't have needed anything from the config file).

The above is a bit convoluted at the moment and there may be a better way to do this, but this should allow the tests to run locally, as well as remotely without touching the real logs. Let me know if I've missed something though or you run in to difficulties.

WilliamMcCumstie:
So it doesn't look like any of the tests are logging to the actual logs. So I haven't gone through and added to `spec` to use the `unit-test` config. 

I will however log it as an issue as it probably will need to be done at some point

BobWhitelock:
This LGTM now.

PR #57, opened 2017-05-24
Templater logs warnings when it tries to use missing parameters

WilliamMcCumstie:
Carry on from the work on the logger. Please merge #55 before this one.

Modified `MetalLog` to override the base `warn` method. Now `MetalLog.warn` will take a message and log it to both `stderr` and the log file. 

`Templater` uses the `warn` when it tries to use a parameter in the `erb` template that is not defined. This is done by creating a wrapper around both `RecursiveOpenStruct` and `MagicNamespace`. 

ERB now receives a `binding` from `TemplaterRecursiveOpenStruct` which is a wrapper around `RecursiveOpenStruct`. If `RecursiveOpenStruct` return a nil value to the templater, the wrapper log it as a `warning`.

However ERB first quires `TemplaterRecursiveOpenStruct` for the `MagicNamespace` object. Then it quires `MagicNamespace` separately for the exact value (e.g. index). Therefore `MagicNamespace` needs a similar wrapper.

### DIFF
--- a/metalware-new/src/templater.rb
+++ b/metalware-new/src/templater.rb
@@ -39,17 +39,18 @@ module Metalware
     end
 
     def method_missing(s, *a, &b)
-      if s == :[]
-        value = a.map { |v| @wrapped_obj[v] }
-        value = value[0] if value.length == 1
-      else
-        value = @wrapped_obj.send(s)
-      end
+      value = @wrapped_obj.send(s)
       if value.nil? && ! @missing_tags.include?(s)
         @missing_tags.push s
         MetalLog.warn "Missing template parameter: #{s}"
       end
       value
+    end
+
+    def [](a)
+      # ERB expects to be able to index in to the binding passed; this should
+      # function the same as a method call.
+      send(a)
     end
   end
 
--- a/metalware-new/src/templater.rb
+++ b/metalware-new/src/templater.rb
@@ -33,13 +33,17 @@ require 'exceptions'
 
 module Metalware
   class TemplaterRecursiveOpenStruct < RecursiveOpenStruct
-    def get_binding
-      return binding()
+    def initialize(*a)
+      @missing_tags = []
+      super(*a)
     end
 
     def method_missing(s, *a, &b)
       value = super
-      MetalLog.warn "Missing template parameter: #{s}" if value.nil?
+      if value.nil? && ! @missing_tags.include?(s)
+        @missing_tags.push s
+        MetalLog.warn "Missing template parameter: #{s}"
+      end
       value
     end
   end
@@ -103,7 +107,7 @@ module Metalware
     private
 
     def replace_erb(template, template_parameters)
-      parameters_binding = template_parameters.get_binding
+      parameters_binding = template_parameters.instance_eval {binding}
       ERB.new(template).result(parameters_binding)
     end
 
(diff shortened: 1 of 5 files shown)


WilliamMcCumstie:
Hold off merging for a moment. Some of the tests are failing

WilliamMcCumstie:
@BobWhitelock Ready for you to review it

BobWhitelock:
FYI rebased on `develop` to better see what changed.

metalware-new/src/templater.rb
```diff
@@ -29,8 +29,25 @@
...
+
+    def method_missing(s, *a, &b)
+      value = super
+      if value.nil? && ! @missing_tags.include?(s)
+        @missing_tags.push s
+        MetalLog.warn "Missing template parameter: #{s}"
+      end
+      value
+    end
+  end
```
BobWhitelock: I like the overall approach you've used for handling the missing parameters, I think it's simple and works well. One improvement I think we could make would be to use a separate object for doing this, something like this (untested, just an idea):

```ruby
class MissingParametersHandler
  def initialize(wrapped_object)
    @wrapped_object = wrapped_object
  end

  def method_missing(*a)
    @wrapped_object.send(a)
    # ... handle misisng parameters as you're doing ...
  end
end
```

We can then use this object to wrap both the `RecursiveOpenStruct` and the `MagicNamespace` object when these are created.

Advantages of this:

- we only need to have the logic of handling missing parameters in one place, rather than in the two places we have it at the moment; if we ever need to add this handling to additional objects in future we also only need to wrap them in a `MissingParametersHandler` rather than re-adding this behaviour to the new objects themselves;

- this separates out the handling of missing parameters from the other behaviour of the wrapped object, making it clearer why we're doing this;

- it avoids the use of inheritance when handling missing parameters for the `RecursiveOpenStruct` - in general you want to favour [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance#Benefits) unless you have a good reason not to; in particular you also normally want to avoid inheriting from classes you don't own which are not particularly intended to be inherited from, otherwise your code is more likely to break if these are changed in future.

metalware-new/src/templater.rb
```diff
@@ -199,70 +211,93 @@ def method_missing(group_symbol)
...
+      @missing_tags = []
+      @struct = MagicNamespaceStruct.new(*a)
     end
 
-    def hunter
-      if File.exists? Constants::HUNTER_PATH
-        Hashie::Mash.load(Constants::HUNTER_PATH)
+    def method_missing(s, *a, &b)
+      if s == :[]
+        value = (a.length == 1 ? @struct[a[0]] : a.map { |v| @struct[v] })
```
BobWhitelock: It's not clear to me what the bug you're trying to fix here is, can you expand on what this is/when this is caused?

Also, why do we need to special case a one element array here? A one element array can be `map`ped the same as one with two or zero or any other number of elements, unless I'm missing something here.

BobWhitelock:
@WilliamMcCumstie: Thanks for making the changes, this looks good now I've just added a few small tweaks.

In particular would you be able to take a quick look at https://github.com/alces-software/metalware/pull/57/commits/8347da3477ad45fe1140f77471e4fdbc32e981c6? This seems to be a simpler/clearer way to do the same thing (at least, all the tests still pass with this change). As far as I can see we don't need to handle `[]` being called with multiple arguments, we just need to handle indexing the same as a method call. Once you've checked that we can merge this.

WilliamMcCumstie:
@BobWhitelock That changes looks like it should work fine

PR #59, opened 2017-05-30
Suppress output with `--quiet`

WilliamMcCumstie:
The CLI now has a global option, `--quiet` which causes standard error to  be redirected to /dev/null. This does not affect standard error and thus we will still get the warnings. 

Also I have included it as part of the `CLI` itself. This means it will not be available if we want to run the ruby tools directly in the future. 

However this looks to meet the requirements at this point in time.

### DIFF
--- a/metalware-old/lib/ruby/lib/alces/stack/boot/run.rb
+++ b/metalware-old/lib/ruby/lib/alces/stack/boot/run.rb
@@ -141,9 +141,9 @@ module Alces
         rescue StandardError => @e
         ensure
           @e ||= Interrupt
-          STDERR.print "Exiting...."
+          $stderr.print "Exiting...."
           e = teardown(@e)
-          STDERR.puts "Done"
+          $stderr.puts "Done"
           raise e unless e == Interrupt
           Alces::Stack::Log.info "clean exit"
           Kernel.exit(0)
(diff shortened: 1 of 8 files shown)


BobWhitelock:
Sorry for not making this clearer in the original improvements document/the Trello card @WilliamMcCumstie - this is not quite what we intended for this option though I can see how you could think it was from the card.

What we wanted was for `--quiet` to leave `stdout` alone, and just hide any unneccessary output to `stderr`. In practise I think this would just mean that `--quiet` hides any warnings that would normally be output to `stderr`, as I believe the only things we currently send to `stderr` are warnings and errors, and the errors will still need to be shown as they're unrecoverable.

In general `stdout` should be for the expected output from a command, while `stderr` is for any extra info which is not part of the core purpose - errors, warnings, debugging info etc. Following this convention means you should be able to pipe commands and only the `stdout`, the main useful output, is directed, while `stderr` is not, meaning you can easily combine tools together.

In our case most of our commands normally effect things directly on the file system, so there is no useful `stdout` to produce and so we don't produce any. One exception is `render` which outputs to `stdout` so you can pipe or redirect a rendered template anywhere; however using the current `--quiet` option just gives this:

```
[root@controller(dev) metalware]# bin/metal render --quiet /var/lib/metalware/repo/pxelinux/default testnode01
warning: Missing template parameter: firstboot
warning: Missing template parameter: kernelappendoptions
```

which means the command does nothing useful with this option.

BobWhitelock:
LGTM, thanks!

PR #60, opened 2017-05-31
Added bash scripts

WilliamMcCumstie:
I have moved power and console into `metalware-new`. I have it set up they invoke a Bash Command class that pulls the command from `ARGV`. This way we can use the same class to run all bash commands (hopefully). 

Currently the bash command handles its own help and has path hard coded within it. Ideally we would merge this into ruby at a latter date.


BobWhitelock:
LGTM, thanks!

PR #61, opened 2017-05-31
Display the usage if not command input

WilliamMcCumstie:
Simple fix to display help when you run `metal`. 

Was looking to do the same for `repo` but that's much harder as `repo use` and `repo update` are defined separately


BobWhitelock:
LGTM!

PR #62, opened 2017-06-01
Make nodegather.sh available in metalware2.0

WilliamMcCumstie:
Moved nodegather.sh into it's new location in metalware. Updated paths
to the new location of nodegather.sh. We may want to move this
functionality from bash into ruby


BobWhitelock:
LGTM!

PR #63, opened 2017-06-05
Make system files available on Metalware server

BobWhitelock:
Link certain files in to Metalware dir; these can then be retrieved by a request to a `/metalware/system/$file` route on the Metalware deployment server.

Trello: https://trello.com/c/oFRKSD7A/29-have-required-urls-work-on-metalware-deployment-server.


WilliamMcCumstie:
LGTM

PR #65, opened 2017-06-05
Show defaults in help

BobWhitelock:
Based on branch under review in https://github.com/alces-software/metalware/pull/63.

Trello: https://trello.com/c/M7EY4cUF/35-show-default-values-for-options-in-command-help.

### DIFF
--- a/src/defaults.rb
+++ b/src/defaults.rb
@@ -1,0 +1,27 @@
+
+require 'constants'
+require 'recursive-open-struct'
+
+
+module Metalware
+  Defaults = RecursiveOpenStruct.new({
+    build: {
+      kickstart: 'default',
+      pxelinux: 'default',
+    },
+    dhcp: {
+      template: 'default',
+    },
+    hosts: {
+      template: 'default',
+    },
+    hunter: {
+      # XXX Always default interface to Metalware build interface (where
+      # possible)?
+      interface: 'eth0',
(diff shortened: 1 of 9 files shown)


src/defaults.rb
```diff
@@ -4,6 +4,11 @@
 
 
 module Metalware
+  # When new defaults are added here you may also want to add these to the
+  # option's help to be displayed.
+  # TODO: Could have these displayed in help automatically by overriding
```
WilliamMcCumstie: Not sure what you mean buy automatically. From the looks of it, changes to the default file will correspond to changes to help. Unless you are talking about automatically searching for a value under: `default.<action-name>.<option-name>`?

BobWhitelock: What I meant was changing our `Command` class in `CommanderExtensions` so when we call `c.option` in `Cli.rb` the default value, if set, will be automatically included in the help text for the option. But that would be more complicated/magic so I've left it for now.

WilliamMcCumstie: Ok probably fine as it is atm. Possible log that as an issue for the future?

BobWhitelock: Thanks for reviewing - I think I'll just leave the TODO comment for now, it should be clear if we need to make the change in future, and no need to add cards for all potential future improvements :slightly_smiling_face:.

PR #66, opened 2017-06-05
Install missing dependencies

WilliamMcCumstie:
Installing the missing dependencies that is preventing metalware
from installing nicely


BobWhitelock:
LGTM

PR #67, opened 2017-06-06
Get repo from config always

BobWhitelock:
This PR fixes an issue where we would sometimes read the path to the repo from the config and sometimes from a hard-coded constant. The key commit with details is in https://github.com/alces-software/metalware/commit/e35bb3cb7535d9c329f0df332a282f281753026b.

This seems a good idea generally, but was particularly important for another feature I've added, the PR for which I'll base off of this.

### DIFF
--- a/spec/commands/build_spec.rb
+++ b/spec/commands/build_spec.rb
@@ -62,11 +62,13 @@ RSpec.describe Metalware::Commands::Build do
 
     it 'renders default standard templates for given node' do
       expect(Metalware::Templater).to receive(:render_to_file).with(
+        instance_of(Metalware::Config),
         '/var/lib/metalware/repo/kickstart/default',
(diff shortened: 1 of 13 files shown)


WilliamMcCumstie:
This looks good to me. However I am still not convinced about passing around the config. It might lead to the situation in which their is a different config in separate classes (possible during testing). 

With that said, I agree that the only alternative is using a global variable that comes with it's own issues. The config is admittedly one of the few cases when a global variable makes sense. 

So I think leave it how it is in this PR for now. However if we are finding that the config is getting passed around a lot, then maybe we should look at making a global variable class that can be initialised with the config before the tests.

BobWhitelock:
As discussed IRL, passing one config around as needed is probably less error prone + easier to test for now.

PR #68, opened 2017-06-06
Require repo before running repo commands

BobWhitelock:
This PR makes commands which require a repo not run before a repo has been set up, and instead fail fast and descriptively.

Based off https://github.com/alces-software/metalware/pull/67; Trello: https://trello.com/c/TnJw3paI/38-commands-which-require-a-repo-to-work-should-gracefully-fail-before-a-repo-has-been-cloned.


BobWhitelock:
@WilliamMcCumstie: as discussed IRL earlier, good suggestion and I've added another commit to improve this, can you take another look and merge this if that's better?

WilliamMcCumstie:
LGTM

PR #70, opened 2017-06-06
Better handle accessing unset parameters

BobWhitelock:
This PR improves the error shown when an attempt is made to access a property on an unset template parameter; see https://github.com/alces-software/metalware/commit/17fade5c2b2090b657c618589e0283ef2e15ff29 for details.

This PR is a fix for https://trello.com/c/lJuWATLd/51-error-when-running-bin-metal-render-var-lib-metalware-repo-kickstart-default-testnode01---trace; it also addresses https://trello.com/c/5DrQ5WUr/52-make-clear-when-errors-are-template-errors-rather-than-metalware and should resolve that unless we can think of other ways we should make this clear to users.

This PR is based on https://github.com/alces-software/metalware/pull/69.

### DIFF
--- a/src/templater.rb
+++ b/src/templater.rb
@@ -64,7 +64,7 @@ module Metalware
       end
 
       def render_to_stdout(config, template, template_parameters={})
-          puts render(config, template, template_parameters)
+        puts render(config, template, template_parameters)
       end
 
       def render_to_file(config, template, save_file, template_parameters={})
--- a/src/templater.rb
+++ b/src/templater.rb
@@ -119,6 +119,12 @@ module Metalware
     def replace_erb(template, template_parameters)
       parameters_binding = template_parameters.instance_eval {binding}
       ERB.new(template).result(parameters_binding)
+    rescue NoMethodError => e
+      # May be useful to include the name of the unset parameter in this error,
+      # however this is tricky as by the time we attempt to access a method on
+      # it the unset parameter is just `nil` as far as we can see here.
+      raise UnsetParameterAccessError,
+        "Attempted to call method `#{e.name}` of unset template parameter"
     end
 
     def nodename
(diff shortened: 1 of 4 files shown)


src/templater.rb
```diff
@@ -119,6 +119,12 @@ def initialize(msg="Input hash may contain infinitely recursive ERB")
     def replace_erb(template, template_parameters)
       parameters_binding = template_parameters.instance_eval {binding}
       ERB.new(template).result(parameters_binding)
+    rescue NoMethodError => e
+      # May be useful to include the name of the unset parameter in this error,
+      # however this is tricky as by the time we attempt to access a method on
+      # it the unset parameter is just `nil` as far as we can see here.
```
WilliamMcCumstie: Alternatively, we control wrapper around the template parameters, hence how we generate the logging. Maybe we could extend it to be a `RecursiveOpenStruct` and handle it before it gets passed through ERB

BobWhitelock: I did consider something like this and have a quick attempt at wrapping the fetched template parameters the other day, so we can then use the wrapper to know which unset template parameter we're attempting to access.

However the issue I ran in to is that by wrapping the return value we are in particular changing the falsiness of it; this in particular makes rendering this in some tests fail as `nested` is always truthy even if the wrapped value is `nil`: https://github.com/alces-software/metalware/blob/develop/spec/fixtures/template.erb#L6. And there seems to be no way of modifying the truthiness of Ruby objects (see https://stackoverflow.com/questions/17988923/make-a-nullobject-evaluate-to-falsy-in-ruby).

I think the alternative approach would be to track all of the unset parameters as we're accessing them so we can know when we're then attempting to access a property of them, but this then adds complexity and room for mistakes in doing this. Happy for you to take a look and see if you can improve this though if you like :slightly_smiling_face:.

WilliamMcCumstie: Leave it as the TODO for now as it adds the functionality. Will create a card in trello so we can come back to this.

PR #72, opened 2017-06-08
Refactored Status to work with metalware v2

WilliamMcCumstie:
Based on PR #71

Continuing on the previous PR, this PR splits the slow tests for status into two files. The first is the monitor tests and the second is the job tests. The Monitor and Job classes have been tested specifically as they handle the multi threading and I want to ensure (well try to) they run correctly.

With that said I have only tested this on a Vagrant VM. So the ping and power functions may not work. However they are based on the same methods as before. The only difference in the Job class is that it calls the power script directly instead of re-invoking metal. 

The Monitor class has been altered slightly so it now passes errors to the main thread when it is joined. Previously it handled and logged errors separately which is redundant as the BaseCommand can do it better. 

The CLI has had a minor bug fix where it would through an error when a value was entered with the optional limits (e.g. `--thread-limit 50` would error).

The changes to the Status Command include:
1. Updating to the iterator to get the nodes list and converting it to an array,
1. Adding parameter validation to preventing the user from putting negative limits in, and
1. When reading completed data, only wait if no data was available on the previous read. This speeds up the case when multiple data reads are required. 

Unrelated to Status:
1. Also fixed a bug in Timeout::Error. The previously used version was depreciated in some of the tests.
1. Also relabeled Bash as BashCommand to make it consistent with BaseCommand (and because `require bash`) is confusing.


BobWhitelock:
This LGTM, though one minor thing is it appears you've included a commit twice while rebasing (https://github.com/alces-software/metalware/pull/72/commits/ff5557348e14dfa143b21e7a6f0729a0b30abb49 and then https://github.com/alces-software/metalware/pull/72/commits/0c532acfe06374a8bc7e53c594c8c59487cb506f), and you then need to remove the duplicated code (https://github.com/alces-software/metalware/pull/72/commits/e898de50986b4e760b6ec32d4e9cd3ee5eb2c2c8#diff-1e4bad586cd5d73944d5a6aa25718011L133). Would it be possible to squash those last two commits so we don't have that?

WilliamMcCumstie:
Ok I have rebased to removed the duplicate commit. Should be ready to merge

BobWhitelock:
Thanks!

PR #73, opened 2017-06-08
Migrated ipmi to metalware version 2

WilliamMcCumstie:
Based on #72 as I needed the new name of BashCommand

The ipmi script has been moved into libexec is being called through the
BashCommand. However there is a clash between Metalware CLI and ipmi's.
It might be worth while rewriting ipmi so that it is ran through ruby
instead of exec.

Also we should consider only using long option tag (e.g. `--opt-tag`)
and removing `-c <config>` tag. This is to prevent conflicts as ruby
will assume you are referring to both simultaneously with a undefined
result. In the meantime, I have resolved a conflict by changing the
ipmi command tag from `-c` to `-k`.

Also commented out the dummy examples in the CLI so they don't get
rendered.

### DIFF
--- a/libexec/ipmi
+++ b/libexec/ipmi
@@ -78,10 +78,10 @@ usage()
 {
   echo "(c) 2008-2011 Alces Software Ltd & Stephen F Norledge"
   echo "alces-ipmi command, perform ipmi commands on single or multiple machines"
-  echo "usage: alces ipmi [-g] <name> -c <command>"
+  echo "usage: alces ipmi [-g] <name> -k <command>"
   echo "  options:"
   echo "    -g specify genders node group"
-  echo "    -c specify ipmi command"
+  echo "    -k specify ipmi command"
   echo "  parameters:"
   echo "    <name> node name or genders group name"
   echo
@@ -93,7 +93,7 @@ while ! [ -z "${*}" ]; do
       NODEGROUP="${2}"
       shift 1
       ;;
-    '-c')
+    '-k')
       CMD="${2}"
       shift 1
       ;;
--- a/src/cli.rb
+++ b/src/cli.rb
@@ -37,7 +37,7 @@ module Metalware
         c.syntax = 'metal repo use REPO_URL [options]'
         c.summary = ''
         c.description = ''
-        c.example 'description', 'command example'
+        #c.example 'description', 'command example'
         c.option '-f', '--force',
           'Force use of a new repo even if local changes have been made to the current repo'
         c.action Commands::Repo::Use
@@ -47,7 +47,7 @@ module Metalware
         c.syntax = 'metal repo update [options]'
         c.summary = ''
         c.description = ''
-        c.example 'description', 'command example'
+        #c.example 'description', 'command example'
         c.option '-f', '--force',
           'Force update even if local changes have been made to the repo'
         c.action Commands::Repo::Update
@@ -57,14 +57,14 @@ module Metalware
         c.syntax = 'metal render TEMPLATE [NODE] [options]'
         c.summary = ''
         c.description = ''
-        c.example 'description', 'command example'
+        #c.example 'description', 'command example'
         c.action Commands::Render
       end
 
       command :hosts do |c|
         c.syntax = 'metal hosts NODE_IDENTIFIER [options]'
         c.summary = ''
-        c.description = ''
+        #c.description = ''
         c.example 'description', 'command example'
         c.option '-g', '--group', String,
           'Switch NODE_IDENTIFIER to specify a gender group rather than a single node'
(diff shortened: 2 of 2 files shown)


libexec/ipmi
```diff
@@ -78,10 +78,10 @@ usage()
 {
   echo "(c) 2008-2011 Alces Software Ltd & Stephen F Norledge"
   echo "alces-ipmi command, perform ipmi commands on single or multiple machines"
-  echo "usage: alces ipmi [-g] <name> -c <command>"
```
BobWhitelock: Do we need any of this `usage` now - won't any  issue with using the `ipmi` command be handled by the Metalware CLI and so this will never be displayed? If so can we remove it?

src/cli.rb
```diff
@@ -150,6 +150,19 @@ def run
         c.action Commands::Status
       end
 
+      command :ipmi do |c|
+        c.syntax = 'metal ipmi NODE_IDENTIFIER [options]'
+        c.summary = 'Perform ipmi commands on single or multiple machines'
+        c.description = "***VOLATILE***\n\n" \
```
BobWhitelock: Can we make this consistent with the other commands marked as 'Volatile'?

src/cli.rb
```diff
@@ -150,6 +150,19 @@ def run
...
+                    'Perform ipmi commands on single or multiple machines'
+        #c.example 'description', 'command example'
+        c.option '-g',
+          'Specifies that NODE_IDENTIFIER is the group. MUST be before NODE_IDENTIFIER'
+        c.option '-k COMMAND',
```
BobWhitelock: Since we always need to specify a command for `metal ipmi`, it seems a nicer syntax for the command would be `metal ipmi NODE_IDENTIFIER IPMI_COMMAND [options]` - how complicated a change do you think this would be? If this is simple to do now without re-writing the whole `ipmi command` in Ruby then we could do this now, otherwise can just add a card for it.

src/cli.rb
```diff
@@ -150,6 +150,19 @@ def run
...
       end
 
+      command :ipmi do |c|
+        c.syntax = 'metal ipmi NODE_IDENTIFIER [options]'
+        c.summary = 'Perform ipmi commands on single or multiple machines'
+        c.description = "***VOLATILE***\n\n" \
+                    'Perform ipmi commands on single or multiple machines'
+        #c.example 'description', 'command example'
+        c.option '-g',
+          'Specifies that NODE_IDENTIFIER is the group. MUST be before NODE_IDENTIFIER'
```
BobWhitelock: Having to specify `-g` before the `NODE_IDENTIFIER` seems inconsistent with our other commands, how complicated do you think it would be to change to allow it anywhere? If this is simple could do this now, otherwise just add a card for it.

libexec/ipmi
WilliamMcCumstie: Their are still a few edge cases in which the bash version of the usage will be displayed. The only way I could see to get around it is to rewrite it in ruby

BobWhitelock: OK thanks for explaining, might as well leave it for now then.

BobWhitelock:
@WilliamMcCumstie: I know we're wrapping the Bash command for now, by changing the CLI syntax I meant would this be simple to do within the Bash script? If not let's just leave it for a potential rewrite.

Also regarding the card for that, we have one here already: https://trello.com/c/apYPki8o/48-better-integrate-wrapped-metalware-commands-for-new-metalware, but you could add any things you think we should consider for the rewrite as comments to that if you like.

src/cli.rb
WilliamMcCumstie: I have updated it so it says `volatile` when you run `metal --help`. The other texts I was going to keep as eventually it will be in the CLI

src/cli.rb
WilliamMcCumstie: It isn't worth changing this in bash because it is probably faster to rewrite in ruby

WilliamMcCumstie:
So most of the ipmi script handles input from the command line, default parameters and the usage. All three of which handled one way or another in Ruby. The only bit of the script we will probably need to keep is: 

```
NO_SLEEP=1
for NODE in $NODES; do
  if [ NO_SLEEP != 1 ]; then
    sleep $STEP_INTERVAL
  fi
  NO_SLEEP=0
  cmd="ipmitool -H ${IPMI_HEAD}${NODE}${IPMI_TAIL} -U${USERNAME} -E $CMD -I lanplus"
  echo -n "Performing command on $NODE.."
  if [ $OUTPUT -eq 0 ]; then
    $cmd > /dev/null 2>&1 && echo OK || echo FAIL
  else
    $cmd || echo FAIL
  fi
done
```

It isn't really worth adjusting the CLI in bash if we are going to scrap it anyway. 

On a slightly different note, so we want to send the following line through the templater when we refactor? That way it's behaviour would be more metalware-ish.

`  cmd="ipmitool -H ${IPMI_HEAD}${NODE}${IPMI_TAIL} -U${USERNAME} -E $CMD -I lanplus"`

BobWhitelock:
LGTM.

> On a slightly different note, do we want to send the following line through the templater when we refactor? That way it's behaviour would be more metalware-ish.

Sounds reasonable

PR #74, opened 2017-06-12
Patch Commander Global Options Bug

WilliamMcCumstie:
Their is a bug in the global options parser in commander. For all global options, the options parser will implicitly generate a short tag based on the first letter (e.g. `--strict` => `-s`). However this short option does not appear in the help.

To make matters more complicated, the short tag does not always work. It only works when their is a specific option in a command with the same short tag (this maybe because of `CommanderExtentions`). As a result the global options parser is leading to weird hard to track bugs. 

To fix them, a fork of the Commander library has been created which will contain the patch. This PR handles the migration to the forked library (url bellow).
https://github.com/alces-software/commander


WilliamMcCumstie:
Please note, at this point in time, I have not finished patching commander. This only points to the new library

WilliamMcCumstie:
`Commander` now contains the `GlobalOption` patch

BobWhitelock:
LGTM; squashed and merged the commits to avoid one commit adding loads of stuff which is removed in the next one.

PR #75, opened 2017-06-13
Each command

WilliamMcCumstie:
I have implemented the each command fully in ruby based on the `Nodes` and `Templater` classes. The old bash method in metalware v1 can be discarded. 

The new method renders the command like any other template and then makes a system call to run it. In order to render the template files, I a new `render_from_string` method has been added to the `Templater`. It is essentially the same as `render` without the file opening component. 

Whilst creating the test suite I had a few issues capturing standardout. I think it is because of how `system` uses `System.spawn` it prints to where STDOUT points and not $stdout. For this reason I needed to add a bit of logic to allow `Each` to redirect to a different fileno (as specified from $stdout and $stderr). This shouldn't affect redirection on the command line as $stdout.fileno will already reflect the new fileno which will be used.

A future improvement maybe to allow `Each` render and run template files.

NOTE: ignore the name of the branch, `Render` has not been modified


BobWhitelock:
LGTM, thanks

PR #77, opened 2017-06-14
Allow iterating through template parameters

BobWhitelock:
When writing Metalware templates it is sometimes desirable to be able to iterate through a set of key-value pairs, rather than needing to access parameters individually; previously this was not possible (without
writing a convoluted template).

This commit allows this by adding a new class to be used for the template parameters loaded from the repo config, which functions the same as `RecursiveOpenStruct` but with an additional `each` method.

This seemed the simplest way to add this functionality without making future issues and confusion likely; we do not want to make this class have all `Enumerable` methods otherwise issues would occur when setting any properties with the same names in the config.

### DIFF
--- a/src/iterable_recursive_open_struct.rb
+++ b/src/iterable_recursive_open_struct.rb
@@ -1,0 +1,26 @@
+
+module Metalware
+  class IterableRecursiveOpenStruct < RecursiveOpenStruct
+    def each(&block)
+      convert_hash_values_to_own_class.each(&block)
+    end
+
+    def each=(*args)
+      raise IterableRecursiveOpenStructPropertyError,
+        "Cannot set property 'each', reserved to use for iteration"
+    end
+
+    private
+
+    def convert_hash_values_to_own_class
+      to_h.map do |k, v|
+        case v
+        when Hash
+          [k, IterableRecursiveOpenStruct.new(v)]
+        else
+          [k, v]
+        end
+      end.to_h
+    end
+  end
+end
--- a/spec/iterable_recursive_open_struct_spec.rb
+++ b/spec/iterable_recursive_open_struct_spec.rb
@@ -1,0 +1,51 @@
+
+require 'iterable_recursive_open_struct'
+
+RSpec.describe Metalware::IterableRecursiveOpenStruct do
+  subject {
+    Metalware::IterableRecursiveOpenStruct.new({
+      prop: 'value',
+      nested: {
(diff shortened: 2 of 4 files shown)


src/iterable_recursive_open_struct.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    def each=(*args)
+      raise IterableRecursiveOpenStructPropertyError,
+        "Cannot set property 'each', reserved to use for iteration"
+    end
+
+    private
+
+    def convert_hash_values_to_own_class
```
WilliamMcCumstie: I know it is unlikely this will happen as the method is a long name. But we should preface convert_hash with a _. This would make it `_convert_hash_values_to_own_class`. Alternatively set a `convert_hash_values_to_own_class=` method that explicitly prohibits it from being overloaded.

BobWhitelock: Not sure what you mean - it's private so shouldn't conflict with anything if someone does choose to use this name I think? E.g.:

```ruby
[25] pry(main)> a = Metalware::IterableRecursiveOpenStruct.new
=> #<Metalware::IterableRecursiveOpenStruct>
[26] pry(main)> a.convert_hash_values_to_own_class
=> nil
[27] pry(main)> a.convert_hash_values_to_own_class = 'foo'
=> "foo"
[28] pry(main)> a.convert_hash_values_to_own_class
=> "foo"
```

WilliamMcCumstie: But would happen if you tried to call `convert_hash_values_to_own_class` from inside `IterableRecursiveOpenStruct`?

Would it still call the private method? or the newly defined value of `convert_hash_values_to_own_class` and return foo?

BobWhitelock: By inside `IterableRecursiveOpenStruct` do you mean within another method in an instance of the class? In that case we want to be able to call it, otherwise it's effectively useless. If not would you be able to write an example with the issue you mean so I can see what we might want to change?

WilliamMcCumstie: Actually it probably wouldn't matter. If a method was to call `convert_hash_values_to_own_class` it would find the private method. However from outside, it would got to `method_missing`.

So LGTM then

BobWhitelock: Thanks for checking, I'll merge this then.

PR #79, opened 2017-06-15
Deep merge template paramter hashes

WilliamMcCumstie:
Previously only the top level hash in the template parameter configs where being merged. Changes in sub-hashes would override the existing sub-hash in its entirety. Now a deep merge occurs where all sub-hashes are also merged.

Also fixed a bug in which only the top level keys where being converted to symbols.


WilliamMcCumstie:
Also removed the last reference to `nodeattr` in `Templater`

WilliamMcCumstie:
These changes LGTM

BobWhitelock:
@WilliamMcCumstie: this LGTM, I've just made a few refactorings and  fixed how we handle nodes not in genders, can you take a quick look at the latest changes and merge them through if you're happy with them?

WilliamMcCumstie:
LGTM, just updated an error message

PR #80, opened 2017-06-16
Remove references to iterator and the source code

WilliamMcCumstie:
Iterator has now been fully replaced with Nodes. As Iterator is no longer been used, it should not be required and the source code removed.


BobWhitelock:
LGTM!

PR #81, opened 2017-06-16
Remove old version of metalware

WilliamMcCumstie:
Now that all the sripts and commands have been migrated into the
new version of metalware, the old version is obsolute. This commit
removes it from the code base. If at a latter date we discover we
need it, we can revert back to this commit.


bin/autocomplete
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env ruby
+
+class AutoComplete
+  def initialize(input = ARGV[0])
+    h = eval(input)
```
BobWhitelock: Do we need `eval` here? I'm not quite sure how this works, would you be able to explain what you're trying to do?

Also this appears to not work for several cases I've tried, e.g.

- `metal "foo.bar<tab>` gives
```
[root@controller(dev) metalware]# metal "foo.barbin/autocomplete:5:in `eval': (eval):1: syntax error, unexpected tIDENTIFIER, expecting '}' (SyntaxError)
"metal", complete: "metal "foo.bar"}
                              ^
(eval):1: unterminated string meets end of file
        from bin/autocomplete:5:in `initialize'
        from bin/autocomplete:44:in `new'
        from bin/autocomplete:44:in `<main>'
```

- in `/tmp/metalware`, `bin/metal <tab>` gives

```
[root@controller(dev) metalware]# bin/metal bin/autocomplete:23:in `add_commands': undefined method `join' for nil:NilClass (NoMethodError)
        from bin/autocomplete:13:in `run'
        from bin/autocomplete:44:in `<main>'
```

WilliamMcCumstie: Sorry that commit should have been on a different branch to this one. Removed it now.

But in relation to your question, it converts the string to a ruby `hash`. Will look into those errors in the current branch on working on it

BobWhitelock:
OK, thanks

PR #82, opened 2017-06-19
Add Metalware design docs, in particular for topologies

BobWhitelock:
The important doc related to topologies is https://github.com/alces-software/metalware/compare/dev/docs?expand=1#diff-7633f31d52d6afaac7f6adf486a7e485.

### DIFF
--- a/docs/design/02-topologies.md
+++ b/docs/design/02-topologies.md
@@ -1,0 +1,288 @@
+
+# Overview
+
+- Will have new `metal configure` command group for configuring values to be
+  used within config files for all nodes, a group of nodes, or an individual
+  node:
+  - `metal configure all`
+  - `metal configure group GROUP`
+  - `metal configure node NODE`
+
+- These commands may also perform other behaviour specific to the type of
+  resource being configured.
+
+- Note that these commands will be just one interface for configuring the
+  'topology' of a cluster; it should be possible for alternative interfaces,
+  such as a future web interface or manual editing, to modify the same files,
+  and the loading of these files for use when templating will be independent of
+  the `configure` commands.
+
+- There will be a `configure.yaml` file in the repo root, specifying
+  configuration questions to be used within these commands or alternative
+  interfaces.
+
+- The `configure` commands will save the entered answers to a file in
+  `/var/lib/metalware/config`, specific to the resource being configured.
+
+- Repo configs will be able to use a new object in the magic `alces` namespace,
+  `alces.config.$question_identifier`, where `$question_identifier` is the
+  identifier for a particular question specified in `configure.yaml`; when
+  templating the corresponding question answer will then be used in these
+  places.
+
+
+# `configure.yaml` format
+
+There will also be a `configure.yaml` in the repo root, which specifies
+configuration questions for the different categories, with the following
+suggested format:
+
+```yaml
+questions:
+  example_question_identifier: &example_question_identifier
+    question: 'What value should this take?' # Required.
+    choices: # Optional; limit choices to one of these options.
+      - 7
+      - 11
+    default: 11 # Optional; default value to use when no input entered, if any.
+    type: 'integer' # Optional/default `'string'`. For validation and converting; see below for details.
+    required: true # Optional/default `true`
+
+  another_example: &another_example
+    question: 'Enter anything?'
+
+all:
+  # Both these questions will be asked for `metal configure all`.
+  example_question_identifier: *example_question_identifier
+  another_example: *another_example
+
+group:
+  # Just ask this question for `metal configure group`.
+  another_example: *another_example
+
+node:
+  # Similarly, just ask this for `metal configure node`. Note: the answer will
+  # be saved under `different_name` in the resulting config this time however.
+  different_name: *example_question_identifier
+```
+
+## Notes regarding above format
+
+- By defining all the questions within a `questions` object we can then share
+  questions between different categories without duplication using YAML's
+  ability to refer to other parts of the document.
+
+- Any name key can be used when specifying a question for a particular
+  category. Often the same identifier will be used for the question definition
+  and when specifying a question for a category, e.g. a `cluster_name` question
+  may be defined in `questions` and then used as `cluster_name` under `all`,
+  however this will not necessarily always need to be the case.
+
+## Notes on `type`
+
+The `type` for a question will serve the dual purpose of validating input and
+then converting this before saving.
+
+For the minimal version of this feature possible type values will probably just
+include `string`, `integer`, and `boolean`. These will be necessary as since we
+are accepting user input we won't know what type we should save a given answer
+as, and we will need non-string types when templating.
+
+Later we can add types with more validation here, e.g. we can have an `ip` type
+which will save the value as a string but validate that it is a valid IP first.
+While these more complicated validations could be specified separately by
+adding an optional regular expression field to question configs, I think
+building these in to Metalware would be better since:
+
+- there will only be a limited number of these more advanced types, and I don't
+  think we need the flexibility of allowing more validations to be specified in
+  the repo;
+
+- having these built in to Metalware means we won't need to duplicate
+  validations between repos or places within repos;
+
+- we may later want to do non-regex validation, e.g. a `dns` type could
+  validate that a particular IP entered gives the expected DNS response (though
+  this particular example sounds like overkill).
+
+
+# `metal configure` commands
+
+The `metal configure` commands each take their specified questions and ask
+them, save the results to a particular file within a new
+`/var/lib/metalware/config/` directory, and perform any other actions
+appropriate to the resource being configured.
+
+## `metal configure all`
+
+Asks the `all` questions; the validated, converted answers will be saved to
+`/var/lib/metalware/config/all.yaml` in the following format:
+
+```yaml
+answers:
+  example_question_identifier: 7
+  another_example: 'foo'
+```
+
+## `metal configure group $GROUP`
+
+Asks for a genders line to be entered first, then similarly to `all` asks the
+`group` questions; the results will then be saved to
+`/var/lib/metalware/config/groups/$GROUP.yaml`
+
+```yaml
+genders: 'somenodes[01-03] somenodes,cluster,all'
+answers:
+  another_example: 'foo'
+```
+
+Note that initially the genders line will just take a line of text, later we
+can add more validation and potentially generate this based on e.g. entering a
+number of nodes in the group and a group name.
+
+Every time `metal configure group` is run we should regenerate
+`/opt/metalware/etc/genders` by (for now at least) simply concatenating all the
+group gender lines together; this way it should always be in sync with the
+configured groups. Aside: it may be worth including a line in the generated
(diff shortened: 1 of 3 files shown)


docs/design/02-topologies.md
```diff
@@ -0,0 +1,288 @@
...
+
+- There will be a `configure.yaml` file in the repo root, specifying
+  configuration questions to be used within these commands or alternative
+  interfaces.
+
+- The `configure` commands will save the entered answers to a file in
+  `/var/lib/metalware/config`, specific to the resource being configured.
+
+- Repo configs will be able to use a new object in the magic `alces` namespace,
+  `alces.config.$question_identifier`, where `$question_identifier` is the
```
WilliamMcCumstie: I don't know if we will want to cal this a `config`. I feel we are over using the term for vastly different things (metalware config, template parameter configs and now typology). Could we possible use a more descriptive name? Or alternatively rename some of the other config files

docs/design/02-topologies.md
```diff
@@ -0,0 +1,288 @@
...
+For the minimal version of this feature possible type values will probably just
+include `string`, `integer`, and `boolean`. These will be necessary as since we
+are accepting user input we won't know what type we should save a given answer
+as, and we will need non-string types when templating.
+
+Later we can add types with more validation here, e.g. we can have an `ip` type
+which will save the value as a string but validate that it is a valid IP first.
+While these more complicated validations could be specified separately by
+adding an optional regular expression field to question configs, I think
+building these in to Metalware would be better since:
```
WilliamMcCumstie: Once we can initialize a validator object with the `yaml` file. Because it is likely the user will make a mistake entering an input. The validator will be responsible for confirming the input and prompting the user to reenter it if an error occurs.

docs/design/02-topologies.md
```diff
@@ -0,0 +1,288 @@
...
+
+
+# `metal configure` commands
+
+The `metal configure` commands each take their specified questions and ask
+them, save the results to a particular file within a new
+`/var/lib/metalware/config/` directory, and perform any other actions
+appropriate to the resource being configured.
+
+## `metal configure all`
```
WilliamMcCumstie: Does the `all` questions get asked when I run `metal configure group/node`? All only when I run `metal configure all`?

docs/design/02-topologies.md
```diff
@@ -0,0 +1,288 @@
...
+
+- `metal configure all` run, an answer `'foo'` is given for the `some_question`
+  question;
+
+- the question `some_question` for `all` is removed from
+  `/var/lib/metalware/repo/configure.yaml`;
+
+- `metal configure all` is run again; `some_question` won't be asked and we
+  should make sure the previous answer is not still included in the resulting
+  config file.
```
WilliamMcCumstie: We should included a `reset` command that deletes `/var/lib/metalware/repo/configure.yaml`. Also we shouldn't alter the main `configure.yaml` file with the answers. We should save the answers in a separate file entirely. That way if we need to reset, we can delete the answer file and everything will be like it is brand new. 

We can still load both files to ensure we don't repeat questions that have all ready been asked.

docs/design/02-topologies.md
```diff
@@ -0,0 +1,288 @@
...
+situations where we want to treat groups or nodes differently.
+
+
+## Command ordering
+
+With this new functionality certain commands will now require that other
+commands have been run first:
+
+- All `metal configure` commands will require the repo be set up (`repo use`
+  has been run) before running.
```
WilliamMcCumstie: Make a note that we have the functionality to enforce this built into the `BaseCommand`

docs/design/02-topologies.md
```diff
@@ -0,0 +1,288 @@
...
+
+- `metal build` will require that `metal configure all` has been run first,
+  i.e. that `/var/lib/metalware/config/all.yaml` has been generated, so any
+  necessary config has been set up before templating.
```
WilliamMcCumstie: Did this previously need the `repo`? Not sure if we should adjust the `BaseCommand` or make a stand alone class to handle it. 

Imo, we should make a separate `Dependencies` class that contains the logic to check if the `repo` or `configure` has been ran. Then we can have a `dependencies` method in the `BaseCommand`. By default the `dependencies` method does nothing as the `BaseCommand` doesn't have any, but we can overload it in the inherited classes.

docs/design/02-topologies.md
```diff
@@ -0,0 +1,288 @@
...
+if a particular config question is optional and nothing has been entered a
+value of `nil` should still be saved, in which case `nil` will be used when
+templating - however if the specified identifier is not present at all then
+this indicates the repo has an inconsistent `config` directory and
+`configure.yaml` file, which should be addressed.
+
+
+## Questions
+
+- May want to change `all` to `cluster` everywhere?
```
WilliamMcCumstie: That might be a good idea if the `all` questions do not get ran with the `group` or `node` questions. `all` implies they do but `cluster` not necessarily.

ste78: I'd originally chosen all to try and avoid people thinking about platforms metalware sets up as single 'clusters', it can currently build an environment that would run several flight clusters for example. Appreciate 'all' might be a little too generic, maybe domain.yaml is a better word than cluster?

docs/design/02-topologies.md
BobWhitelock: Yeah I agree, added something briefly about that here: https://github.com/alces-software/metalware/pull/82/files/395b7823017437099db9a3b9fb7f60fdb11b9512#diff-7633f31d52d6afaac7f6adf486a7e485R286. Not sure what the best terms to use might be, any ideas?

@ste78 would you describe the 'topology' of the cluster as being what we currently call the 'config' in the repo, or the new planned config files for `all`, `groups` and `nodes` - or the combination of both of these? Wondering if one of these should be renamed to `topology` for clarity.

docs/design/02-topologies.md
BobWhitelock: Not quite sure what you mean? I think we may be thinking similar things; here I just mean that the types available to be specified in these files will be hard-coded in to Metalware, and we will need some kind of object to validate these which we can then use from both `configure` commands and a future web UI.

docs/design/02-topologies.md
BobWhitelock: Nope, the plan is that only the questions specified for each level in `configure.yaml` will be asked when `configure` is run for that level. So some questions may be asked at multiple levels but only those specified. This is because some things it only makes sense to ask at a particular level, e.g. specifying `cluster_name` at a level other than `all` is only likely to make things confusing and broken. Other things people might want to specify differently at different levels for odd use cases, e.g. a DNS server to use, but asking that question for every node you want to configure is probably overkill for the default topology, so we'll probably leave that at the `all` level by default.

docs/design/02-topologies.md
```diff
@@ -0,0 +1,288 @@
...
+Note that initially the genders line will just take a line of text, later we
+can add more validation and potentially generate this based on e.g. entering a
+number of nodes in the group and a group name.
+
+Every time `metal configure group` is run we should regenerate
+`/opt/metalware/etc/genders` by (for now at least) simply concatenating all the
+group gender lines together; this way it should always be in sync with the
+configured groups. Aside: it may be worth including a line in the generated
+genders indicating this and that any changes to it directly may not persist.
+
```
BobWhitelock: Sure we can do that, and we definitely will want to if we add specifying genders with more structure later. It will basically amount to the same thing whether we just concatenate the lines together directly or iterate through the groups in a template and include their genders lines, I was just thinking the former would be marginally quicker for the MVP.

docs/design/02-topologies.md
BobWhitelock: Regarding not altering `configure.yaml`, that's the plan, the answers will be saved to files in  `/var/lib/metalware/config/` (as described above, sorry if that's unclear). The directory used within `/var/lib/metalware` used may be named differently depending on the discussion in https://github.com/alces-software/metalware/pull/82/files/395b7823017437099db9a3b9fb7f60fdb11b9512#r122666883, but the same file structure within this as described above should be used.

Regarding a `reset` command - I think with the Metalware CLI, since any user should have full interactive root access to the machine with Metalware on, adding commands which simply duplicate shell commands seems unnecessary. `reset` would just be the same as `rm`, or `rm -rf /var/lib/metalware/config` to remove all configuration, so I'm not sure we need it. @ste78 what do you think?

docs/design/02-topologies.md
WilliamMcCumstie: Sorry my bad. This was just my initial musings on the topic. Have quite get it as a fully coherent thought

BobWhitelock: No problem! It's good to clarify things :slightly_smiling_face:

docs/design/02-topologies.md
BobWhitelock: Thanks, about to push a commit which does this.

docs/design/02-topologies.md
BobWhitelock: Yep it did.

What you describe sounds like a possible good way to generalise specifying dependencies.

We could also consider specifying what files a command depends on in the new `yaml` file specifying the CLI you're working on, and we can then just have standard logic in base command to check each required file exists before proceeding.

docs/design/02-topologies.md
BobWhitelock: `domain` sounds like the best choice, will add that in as a change to make.

docs/design/02-topologies.md
WilliamMcCumstie: Yeah sure, should be easy enough to include into the YAML. We may need to make a change to commander extensions however

WilliamMcCumstie: It might be hard to define it in the cli yaml config. Think of it as how `Commander` behaves and not how `BaseCommand` works. It's probably just easier to define a `DEPENDENCIES_LIST` in the inherited classes. Then in the `BaseCommand` check if the `DEP` constant exists. If it doen't then their is no dependencies. If it does then enforce them.

BobWhitelock: Yep fair enough, that sounds like a good plan - I'll amend my last commit to add a note about this possible implementation.

docs/design/02-topologies.md
ste78: topology i think is the combination of both..

How about `/var/lib/metalware/answers` & `alces.answers.$question` ?

BobWhitelock: @ste78 - yep that may be the best choice, I'll add to my last commit to mention that in the doc

BobWhitelock:
@WilliamMcCumstie: In the topologies doc I mention that we should first accept a `genders` line for `configure group`, and save this in the `answers` file. Just to let you know for implementation purposes, after thinking more I think it is probably unnecessary to do this, we can just accept this as any other answer and save it in the usual way. This means we won't need a special case for groups, and can just save all answers at the top level in the `answers` file.

WilliamMcCumstie:
Ok LGTM then. We should go through the implementation changes required. After you merge this request we should branch of a `release/2.0.0` and polish it off to merge into master. Then the implementation of typologies can be version `2.1.0`

PR #83, opened 2017-06-20
Move cli information into yaml document

WilliamMcCumstie:
As part of completing the autocomplete, the cli information has been moved into a yaml config. The config contains (for the most part) the same header tags as used by `Commander`. These tags are then interpreted by `parser.rb` and applied to the command object.

The yaml also hierarchically defines the using referencing. Their is no need to include the `sub_command` and `sub_command_group` tags into the `yaml` document itself. The parser will automatically infer when the tags are required from the structure. NOTE the sub_command tags are only available in the `alces` version of `Commander`


BobWhitelock:
Looks great, thanks!

PR #84, opened 2017-06-20
Integrate auto-completion in with cli config

WilliamMcCumstie:
This is a continuation on cli `config.yaml`.  Merge #83 first

The auto-completion now loads the cli `config.yaml` file and pulls the commands and the autocomplete from it. `parser.rb` is no longer used by the autocomplete and the code designed to facilitate this has been removed.


BobWhitelock:
Looks like you added everything in this branch to https://github.com/alces-software/metalware/pull/83, so this closed when that was merged - is that correct?

PR #90, opened 2017-06-23
Combined and integrated the topology answers into the magic namespace

WilliamMcCumstie:
First generalised how Node combined it's config files which required splitting `raw_config` into two parts. This half generates a an array of config hashes to be combined, then a combine_hash method merges them into one in order. The answer yaml files are processed in the same manor.

Then to gain access to the answer files in the magic namespace, the `Node` object is passed in instead of the `nodename`. The `Node` object is not passed to the underlining struct so it can not be accessed by `alces.node` from a template. However the answer can be accessed with `alces.answers`.

The answers have been integrated in with the MissingParameterWrapper, so by default they throw an `warning` if they are missing, which can be converted to an error with `--strict`


spec/templater_spec.rb
```diff
@@ -85,7 +85,7 @@ def expect_renders(template_parameters, expected, config: Metalware::Config.new)
 
         expect{
           Metalware::Templater.new(@config)
-        }.to raise_error(Metalware::Templater::LoopErbError)
+        }.to raise_error(Metalware::LoopErbError)
```
WilliamMcCumstie: Will do in the future. Typically tried to, but forgot on the last commit

spec/fakefs_helper.rb
```diff
@@ -0,0 +1,50 @@
...
+# For more information on the Alces Metalware, please visit:
+# https://github.com/alces-software/metalware
+#==============================================================================
+require 'fakefs/safe'
+
+class FakeFSHelper
+  def initialize(metal_config)
+    @metal_config = metal_config
+    FakeFS.deactivate!
+    FakeFS.clear!
```
WilliamMcCumstie: Requiring `FakeFS` is error prone. For example `require 'fakefs` will immediately start faking the FS. That is why `require 'fakefs/safe'` needs to be used instead. 

Also because their can only be one faked `FS`, I am not sure how it will behave between the tests. Also turning off the faked FS doesn't delete it. If you turn it back on, then it will be the same as before.

So the `clear!` insures that the faked FS is deleted (i.e. only has the root dir `/`) and the `deactivate!` insures that it isn't running. These two actions are independent and can happen in any order.

spec/fakefs_helper.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def run(&block)
+    FakeFS do yield end
+  end
```
WilliamMcCumstie: Requiring `FakeFS` is error prone. For example `require 'fakefs` will immediately start faking the FS. That is why `require 'fakefs/safe'` needs to be used instead. 

To prevent this ever becoming an issue, I wanted to abstract away all of the `FakeFS` commands we will use regularly into `FakeFSHelper`. This way we don't need to worry about `requring 'fakefs'` into the test themselves. 

tl;dr; We don't absolutely need it, but it might make our lives easier if we have it anyway

spec/templater_spec.rb
```diff
@@ -134,6 +139,31 @@ def expect_environment_dependent_parameters_present(magic_namespace)
       SpecUtils.use_mock_genders(self)
     end
 
+    context 'with answer files' do
+      it 'loads the answer files' do
```
WilliamMcCumstie: fixed

src/templater.rb
```diff
@@ -228,10 +227,19 @@ def method_missing(group_symbol)
...
+  MagicNamespace = Struct.new(:index, :node, :firstboot, :files) do
+    def initialize(index: 0, node: nil, firstboot: nil, files: nil)
       files = Hashie::Mash.new(files) if files
-      super(index, nodename, firstboot, files)
+      @node = node
+      super(index, nil, firstboot, files)
+    end
+
+    def nodename
+      @node ? @node.name : nil
```
WilliamMcCumstie: fixed

BobWhitelock:
Thanks for the various fixes and explanations. I've made a few additional improvements to some bits, have a look through if you like.

One other point, can you try to avoid merging `develop` in to branches? Generally we try to follow [this](http://nvie.com/posts/a-successful-git-branching-model/) sort of workflow to keep a more consistent and understandable structure of branches.

In this case, where you want to incorporate recent changes on `develop` into an earlier branch off of it, you could either achieve this by rebasing on `develop` or resolving the conflicts when the branch is merged back into it. Rebasing can sometimes make resolving things easier as you can resolve each commit in turn rather than the whole branch at once.

PR #91, opened 2017-06-27
Integrate default answers into typology questions

WilliamMcCumstie:
Continuation on typology answers. Please merge #90 first.

Firstly a few house keeping code changes have been made in this PR. The auxiliary commands (`BaseCommand`, `BashCommand` and `ConfigureCommand`) have been moved into their own module. This is to emphasise that they directly affect the running of the commands without being a command themselves. It also means that all the classes in `Commands` and `CommandHelpers` modules need to inherit from `BaseCommand` (notable exception of `BaseCommand` itself). 

Also simplified the assignment of the Questions object to the questions variable in `Configurator`. Previously it was double assigned during initialize. Now it is not created during initialize, but instead when it is first created. 

Next the default answers where integrated into highline. This required patching highline as it also uses erb. This meant that erb tags in the default answers where being rendered by highline causing an error. The patch escapes them out.

For the patch to become effective, a new copy of `HighLine` needs to be created with the `Configurator` object. Otherwise the additional code in `Configurator` handles the distinction between the different types of default object classes.


src/configurator.rb
```diff
@@ -78,29 +84,54 @@ class Question
...
           questions_section: questions_section
         )
       end
 
+      def add_default(question, type = :string)
+        unless @default.nil?
+          parsed_default = nil
+          case type
+          when :boolean
+              if @defualt == "true" || @defualt.is_a?(TrueClass)
```
BobWhitelock: `@defualt` != `@default`!

Also, I don't think it's the responsibility of setting defaults to try and coerce things to the correct type. It seems like extra code without a real benefit; you should specify the correct type in the `configure.yaml` file, and this will just let people write sloppy `configure.yaml`s.

I think a more correct way to handle the wrong type being used would be to add a step (earlier in `configure`) to validate the `configure.yaml` is the correct format, which would validate `default` as well as `required` are the correct type among other validations.

However since this is written it does provide some value for now, so the best step now may be to add a note to the code and a corresponding card regarding adding validation of `configure.yaml`. Thanks.

spec/configurator_spec.rb
```diff
@@ -270,28 +270,16 @@ def configure_with_input(input_string)
...
 
-      configure_with_input("\n\n\n\n\nyes\nno\n")
+      configure_with_input("\n\n\n\n\n")
 
       expect(answers).to eq({
         'string_q' => str_ans,
         'string_erb' => erb_ans,
-        'integer_q' => 10,
-        'bool_true' => true,
-        'bool_false' => false
```
BobWhitelock: Ignore previous comment about doing this :).

src/configurator.rb
```diff
@@ -65,11 +68,14 @@ def save_answers(answers)
...
-        properties: properties,
-        configure_file: configure_file,
-        questions_section: questions_section
+      Question.new({
+          identifier: identifier,
+          properties: properties,
+          configure_file: configure_file,
+          questions_section: questions_section
+        }.tap { |h|
+          h[:default] = properties["default"] if properties.key?("default")
```
WilliamMcCumstie: Fixed

src/configurator.rb
```diff
@@ -78,29 +84,54 @@ class Question
...
         @choices = properties[:choices]
+        @default = default
         @type = type_for(
           properties[:type],
           configure_file: configure_file,
           questions_section: questions_section
         )
       end
 
+      def add_default(question, type = :string)
```
WilliamMcCumstie: no it does not, fixed

spec/configurator_spec.rb
```diff
@@ -229,5 +248,51 @@ def define_questions(questions_hash)
...
+          },
+          bool_false: {
+            question: 'False?',
+            type: 'boolean',
+            default: false
+          }
+        }
+      })
+
+      configure_with_input("\n\n\n\n\nyes\nno\n")
```
WilliamMcCumstie: So turns out that `booleans` can't have defaults (or I am not doing it correctly). Their is a latter commit that fixes this. But will also fix it now

WilliamMcCumstie: Actually turns out I already did

WilliamMcCumstie:
Made a few minor changes. Should be ready to merge once #90 has been

WilliamMcCumstie:
I've rebased and now a test is broken

WilliamMcCumstie:
Ok looks like a fixed the test somewhere along the line

PR #92, opened 2017-06-27
Define configure default answers as the previously recorded answer

WilliamMcCumstie:
Continuation of Typology questions. Please merge #91 first

When loading the default answer, the answer file is checked first. If an old answer is found, then that is used as the default. Otherwise the default in `configure.yaml` is used.

### DIFF
--- a/spec/configurator_spec.rb
+++ b/spec/configurator_spec.rb
@@ -31,33 +31,44 @@ RSpec.describe Metalware::Configurator do
   }
 
   let :configurator {
+    make_configurator
+  }
+
+  def make_configurator(hl = highline)
     Metalware::Configurator.new(
-      highline: highline,
+      highline: hl,
       configure_file: configure_file_path,
       questions_section: 'test',
       answers_file: answers_file_path
     )
-  }
+  end
 
   def define_questions(questions_hash)
     configure_file.write(questions_hash.to_yaml)
     configure_file.rewind
   end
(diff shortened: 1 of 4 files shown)


spec/configurator_spec.rb
```diff
@@ -282,5 +292,38 @@ def configure_with_input(input_string)
...
+            type: 'integer',
+            default: 10
+          }
+        }
+      })
+
+      old_answers = {
+        "string_q" => "CORRECT",
+        "integer_q" => -100,
+        "should_not_see_me" => "OHHH SNAP"
```
BobWhitelock: :shipit:

BobWhitelock:
Tests are failing for me on this branch. I thought it might be because of my earlier rebase, but they're also failing on your version at https://github.com/alces-software/metalware/tree/feature/rerun-configure-OLD. Note that branch also has an issue as there's a diff marker left in https://github.com/alces-software/metalware/commit/167a4f42767d8ddebc735e2ae5e922bcb010173d, which was a reason I rebased it, so it may be simpler to resolve this off of this branch.

WilliamMcCumstie:
Seems to be an issue with one of the many rebases. The diff marker has been fixed in another rebase. All the test should be passing now (again).

bobwhitelock:
LGTM, just did a little bit of refactoring before merging.

PR #97, opened 2017-06-30
Make node indexes consistent across different commands

BobWhitelock:
This PR makes the node `index` used when templating be consistent across invocations of different commands which template for the same node. For example, given:

```
# genders
node[01-05] nodes,domain
```

then `metal build -g nodes` and `metal build node05` will both use an `index` of `4` for `node05`, whereas previously they would use `4` and `0` respectively.

This does have the limitation described [here](https://github.com/alces-software/metalware/blob/bd781fd6630350344e61c6afa332be2e6e5a95fe/spec/node_spec.rb#L168,L177); however this situation should be improved with future changes, as described in https://trello.com/c/XjtdbEHm/80-improve-how-indexes-work, which this PR is an initial step towards resolving.

### DIFF
--- a/docs/templating-system.md
+++ b/docs/templating-system.md
@@ -63,8 +63,12 @@ Values available within this namespace are:
 <td><code>index</code></td>
 <td>
 
-The index of the current node within the group being templated, for the current
-group of nodes being templated.
+The index of the current node within its primary group. The primary group for a
+node is the first group associated with that node in the genders file, i.e. the
+first group to appear in the output of `nodeattr -l $NODE_NAME`. The
+<code>index</code> is guaranteed to be consistent between invocations of
+different commands which template for the same node, so long as the genders
+file remains consistent.
 
 </td>
 
--- a/spec/commands/build_spec.rb
+++ b/spec/commands/build_spec.rb
@@ -80,7 +80,6 @@ RSpec.describe Metalware::Commands::Build do
     def expected_template_parameters
       {
         nodename: 'testnode01',
-        index: 0,
         firstboot: true,
         files: SpecUtils.create_mock_build_files_hash(self, 'testnode01'),
       }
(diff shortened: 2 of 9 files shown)


WilliamMcCumstie:
P.S. It also means that we don't need to handle the case when a node is part of multiple groups, as it will always determine index of the `PrimaryGroup`

BobWhitelock:
I understand the issues you have raised, and agree that this PR by itself does not give us a perfect solution for how indexes work. However, I don't think your suggestion will resolve this or have `index` work the way we want:

- The 2 issues you mention are already the case, if different `metal` templating commands are run over different groups of nodes - this PR makes the `index` of a node consistent between running different commands over different groups, which may include the same nodes. In order to do this we needed to have a consistent group used for each node, and to that end I introduced the concept of a 'primary' group. However, this PR does nothing to make primary groups easier to work with or to ensure you assign nodes to the 'correct' primary group.

- This would not really address nodes having odd indexes, for example if the entry for `nodes` in the genders file appears after some other groups, then `node01` may have `index` `5`, which also seems un-intuitive.

- Assigning indexes in this way would make generating things like IPs more likely to break. For instance as soon as over 256 nodes are defined in the genders file, bad IPs for nodes after this would be generated where the `index` is used to do this; for instance the 257th node would be given an IP like `10.10.0.256`, which is meaningless.

- I think it adds another point of confusion that you would have to include a magic `PrimaryGroup` group for every group you set up, and if you do not an error may be thrown at some later time when another command is run.

Fundamentally, this would be a change in how `index` operates, making it globally unique amongst all nodes. This would be different to how Steve wanted the `index` to work the other day, as well as how we thought of it operating before (this PR is still a change from that, but it preserves a similar idea of the `index` being for the node within its group, while making this consistent between different commands). I think there may be scope for adding a template parameter for a global index for nodes, if it is requested that this would be useful, but I don't think we should replace the existing concept of `index` to do this.

This is just a step towards improving how indexes and groups work. I'm not certain how this will finally look, but some things I'm considering are:

- When `metal configure group $GROUP` is run, `$GROUP` will be the primary group. Rather than having a full genders line be given when this is run, you will instead be able to enter some additional groups as well as a node specification (e.g. `node0[01-40]`).

- When `metal configure domain` is run, we could also allow specifying additional genders lines in some way; for example with `metal configure group` you could define the primary group of `node0[01-99]` as `nodes`, but here you could specify additional groups like `rack1` for `node0[01-10]`.

- Using the above entered `answers`, we can then regenerate the genders file each time one of these commands is run. The primary groups, with specified nodes and any additional groups, will appear first, while any other lines can be included afterwards. This way the genders file will always be in the format we expect, as well as there being a more defined process for users to modify it in expected ways.

Since you say the code's OK I'll merge this through, but we can discuss the ideas more tomorrow or on here if you like :slightly_smiling_face:.

PR #98, opened 2017-06-30
Created Dependencies class for Repo and Configure

WilliamMcCumstie:
The new dependencies class will enforce that the necessary files in the repo and topology/ configure answers exist. The is a generalised two step process for validating both the repo and the configure answers.

The first step is checking if the base directory exists (set in either the config or constants). The second step is to prepend the base directory to a list of files past in with the dependency hash. `Dependencies` will then check if all these files exist. Note that if `true` is passed in with the dependency hash instead of a list of files, then the second step is skipped. 

Additional logic is included for validating if the base answer directory exists for validating configure. Answer files will only be validated if `configure.yaml` is in the `repo`. This is to allow backwards compatibility, with non-configurable repos. Also it is a convenient way to turn of the check during debugging.


src/dependencies.rb
```diff
@@ -0,0 +1,110 @@
...
+        values.each { |value| validate_dependency(dep, value) }
+      }
+    end
+
+    private
+
+    attr_reader :config, :command
+
+    def validate_dependency(dep, value)
+      return if send(:"validate_#{dep}") == "skip"
```
BobWhitelock: I don't really like relying on a magic string being returned to determine if we don't need to run a particular validation. If we need this, we could have a `should_validate_#{dep}?` method as well as a `validate_#{dep}` method, with the former determining if the validation is needed while the latter just does the actual validation. That would make things clearer and give each method a single purpose.

With that said, I can't really see a need for conditional validations. I think we should just enforce the validations in all cases; otherwise we are making things more complex without a real benefit, as no one is really depending on the previous behaviour in any case.

When it comes time for the next release we can decide if this is a breaking change or not, but we shouldn't make our code over-complicated to avoid having to decide this (and risk more bugs due to the added complexity in the process).

src/dependencies.rb
```diff
@@ -0,0 +1,110 @@
...
+      unless file_exists?(dep, value)
+        raise DependencyFailure, get_generic_failure_message(dep, value)
+      end
+    end
+
+    def validate_repo
+      @validated_repo ||= begin
+        msg = "'#{command}' requires a repo. Please run 'metal repo use'"
+        raise DependencyFailure, msg unless file_exists?(:repo, "", true)
+        true
```
BobWhitelock: Do we need to return `true`? Do we check for this anywhere?

spec/fakefs_helper.rb
```diff
@@ -41,18 +41,30 @@ def add_answer_files(files)
     end
   end
 
+  # Deprecated, please use clone_repo
+  def load_repo(path = @metal_config.repo_path)
+    FakeFS::FileSystem.clone(path, @metal_config.repo_path)
+  end
+
+  # Deprecated, please use clone_repo
+  # TODO: remove references to load_config_files
```
WilliamMcCumstie: fixed

spec/fakefs_helper.rb
```diff
@@ -41,18 +41,30 @@ def add_answer_files(files)
...
   def load_config_files(config_names = [])
     FakeFS::FileSystem.clone(@metal_config.configure_file)
     config_names.each do |c|
       FakeFS::FileSystem.clone(@metal_config.repo_config_path(c))
     end
   end
 
+
```
WilliamMcCumstie: fixed

spec/dependencies_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+
+    it 'fails if the repo doesn\'t exist' do
+      @fshelper.clear
+      expect {
+        run_dependencies(@config, { configure: "domain.yaml" })
+      }.to raise_error(Metalware::DependencyFailure)
+    end
+
+    # NOTE: This is the backwards compatibility test
+    it 'pass if the repo exists but configure.yaml doens\'t' do
```
WilliamMcCumstie: fixed

src/dependencies.rb
WilliamMcCumstie: The purposes of the true is to set a value for `@validate_repo`. Originally this was called multiple times so it allows for the method to be short circuited as it only needs to be checked once. Now that `Dependencies` has been refactored, it doesn't get called as many times. However it is still will be called more than once. 

I have added a comment explaining this in the new version

src/dependencies.rb
WilliamMcCumstie: I have removed the magic parameter and simplified the code. It does mean that it is no longer backwards compatible. This is also why `build` integration test had to be updated.

WilliamMcCumstie:
I have changed the `dependency_hash` to only except an array. `Strings` have be replaced with single element arrays and `true` has been replaced with empty arrays.

spec/dependencies_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+
+require 'fakefs_helper'
+require 'spec_helper'
+require 'fileutils'
+
+RSpec.describe Metalware::Dependencies do
+  before :each do
+    @config = Metalware::Config.new
+    @fshelper = FakeFSHelper.new(@config)
+    @fshelper.load_repo(File.join(FIXTURES_PATH, "repo"))
```
WilliamMcCumstie: fixed

spec/dependencies_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+  end
+
+  def run_dependencies(config, dep = {})
```
WilliamMcCumstie: fixed

spec/dependencies_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+    @fshelper.run do
+      Metalware::Dependencies.new(config, "test", dep).enforce
+    end
+  end
+
+  context 'repo dependencies' do
```
WilliamMcCumstie: fixed

spec/dependencies_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+    it 'fails if the repo doesn\'t exist' do
+      @fshelper.clear
```
WilliamMcCumstie: fixed

spec/dependencies_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+    end
+  end
+
+  context 'configure dependencies' do
+    before :each do
+      @fshelper.clone_answers(File.join(FIXTURES_PATH, "answers/basic_structure"))
+    end
+
+    it 'fails if the repo doesn\'t exist' do
+      @fshelper.clear
```
WilliamMcCumstie: fixed

bobwhitelock:
This looks good, thanks for going back over things :slightly_smiling_face: .

PR #100, opened 2017-07-03
Fix hunter bugs

WilliamMcCumstie:
This PR fixes two bugs that where identified in hunter:

The first bug is when `-s` is used with `hunter`, it truncates the `UDP` packets. This is because `Pcap` always passed the `ARGV` array into `OptionParser`. This has been fixed in Pcap (link below). In short, `Pcap` no longer parses the `ARGV` array unless it isn't initialised with any inputs.
https://github.com/alces-software/ruby-pcap/commit/966ce3739d7c91c15032b3a85095c7400565841f

The second bug is when an integer value is entered into the cli with a leading zero (e.g. `-s 035`). As `Integers` are converted according to their prefixed value, '035' is converted as an octal to 29. Now the `cli_helper` will automatically set all `Integers` as `DecimalIntegers`. A bug patch also needed to be included in `Commander` (link below). This patches the underling ruby function, which doesn't enforce DecimalIntegers. This meant that even `DecimalInteger 035` was being converted to 29. This bug patch fixes that. 
https://github.com/alces-software/commander/commit/e9c34dad0d2688f4c1f8241810fcbf5e1a96cbf5

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -7,14 +7,14 @@ GIT
 
 GIT
   remote: https://github.com/alces-software/ruby-pcap.git
-  revision: 42728a34e955ad00a5d033f4ef999a75a15f6f67
+  revision: 966ce3739d7c91c15032b3a85095c7400565841f
   specs:
-    pcap (0.7.7)
+    pcap (0.7.8)
 
 GEM
   remote: https://rubygems.org/
   specs:
-    activesupport (5.1.1)
+    activesupport (5.1.2)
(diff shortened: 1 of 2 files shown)


WilliamMcCumstie:
Once merged in, this should be labelled as `2.0.0-rc2`

bobwhitelock:
Looks great, thanks for looking into these issues :slightly_smiling_face:.

Will merge then need to make one other quick change before putting out `2.0.0-rc2`.

PR #103, opened 2017-07-06
Enforce required and allow optional questions

WilliamMcCumstie:
Configure questions are now required by default. This means that the
user will be prompted to renter an answer if they do not give one and
it does not have a default.

A new optional tag can be included with the questions, which means
an empty input is a valid response. The optional tag must evaluated
to a truthy value.

Note that when HighLine re-prompts for an input, it uses inspect to
generate the message. This gives crazy responses such as:
`must match <proc:/some/file/path-352random-other-stuff>`

As this isn't very useful to the user, a HighLinePrettyValidateProc
class is used. This class overrides the inspect method so that it
gives nicer prompts to the user.

### DIFF
--- a/spec/configurator_spec.rb
+++ b/spec/configurator_spec.rb
@@ -340,7 +340,15 @@ RSpec.describe Metalware::Configurator do
       })
 
       expect{
-        configure_with_input("\n")
+        old_stderr = STDERR
+        begin
+          $stderr = Tempfile.new
+          STDERR = $stderr
+          configure_with_input("\n")
+        ensure
+          STDERR = old_stderr
+          $stderr = STDERR
+        end
       }.to raise_error(EOFError)
     end
 
--- a/spec/configurator_spec.rb
+++ b/spec/configurator_spec.rb
@@ -344,12 +344,20 @@ RSpec.describe Metalware::Configurator do
         begin
           $stderr = Tempfile.new
           STDERR = $stderr
-          configure_with_input("\n")
+          configure_with_input("\n\n")
         ensure
           STDERR = old_stderr
           $stderr = STDERR
         end
+        # NOTE: EOFError occurs because HighLine is reading from an array of
+        # end-line-characters. However as this is not a valid input it keeps
+        # re-asking until it reaches the end and throws EOFError
       }.to raise_error(EOFError)
+
+      output.rewind
+      # Checks it was re-asked twice.
+      # The '?' is printed when the question is re-asked
+      expect(output.read.scan(/\?/).count).to eq(2)
     end
 
     it 'allows optional questions to have empty answers' do
--- a/spec/configurator_spec.rb
+++ b/spec/configurator_spec.rb
@@ -10,8 +10,12 @@ RSpec.describe Metalware::Configurator do
     Tempfile.new
   }
 
+  let :output {
+    Tempfile.new
+  }
+
   let :highline {
-    HighLine.new(input)
+    HighLine.new(input, output)
   }
 
   let :configure_file {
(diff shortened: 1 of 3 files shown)


bobwhitelock:
(Rebased on `develop` so we have the integration spec fix on this branch)

src/configurator.rb
```diff
@@ -112,7 +110,11 @@ def initialize(identifier:, properties:, configure_file:,
       def ask(highline)
         ask_method = "ask_#{type}_question"
         self.send(ask_method, highline) do |highline_question|
-          highline_question.default = default
+          if !default.to_s.empty?
```
bobwhitelock: This is possibly a roundabout/over-complex way to test that a default has been given - do you think we can just use [`if default.present?`](http://guides.rubyonrails.org/active_support_core_extensions.html#blank-questionmark-and-present-questionmark) and get the same effect?

src/configurator.rb
```diff
@@ -173,7 +175,25 @@ def determine_default(question_default:, old_answer:)
...
+      class HighLinePrettyValidateProc < Proc
+        def initialize(print_message, &b)
+          # NOTE: print_message is prefaced with "must match" when used by
+          # HighLine validate
+          @print_message = print_message
+          super(&b)
+        end
+
+        def inspect
+          @print_message
```
bobwhitelock: I like this solution for working around HighLine giving an odd error message - could just be worth adding a comment about why this is needed since it's not obvious why we need to override `inspect`.

src/configurator.rb
```diff
@@ -173,7 +175,25 @@ def determine_default(question_default:, old_answer:)
           end
         end
       end
-    end
 
+      def required_answers_check
```
bobwhitelock: A better name here might document the intent of what we're doing - possibly something like `ensure_answer_given`?

src/configurator.rb
```diff
@@ -112,7 +110,11 @@ def initialize(identifier:, properties:, configure_file:,
       def ask(highline)
         ask_method = "ask_#{type}_question"
         self.send(ask_method, highline) do |highline_question|
-          highline_question.default = default
+          if !default.to_s.empty?
+            highline_question.default = default
+          elsif !optional
```
bobwhitelock: I think `if !optional` would read clearer as `if required` - you could extract a `required` method which is just defined as `!optional`; or (possibly simpler) instead just save the original field as `@required = !properties[:optional]`, since we don't need to access `optional` anywhere other than here.

bobwhitelock:
Also: looks like this commit has introduced a rogue newline being output, do you know where that's coming from?

<details>

```
➜  metalware git:(feature/optional-answers) make
bundle exec rspec \
        --force-colour \
        --exclude-pattern 'spec/slow/**/*, spec/integration/**/*'

Randomized with seed 23925
....................
........................................*...................................

[...snip...]

➜  metalware git:(feature/optional-answers) gco HEAD~
Note: checking out 'HEAD~'.

[...snip...]

HEAD is now at 12c1376... Merge pull request #98 from alces-software/feature/dependencies
➜  metalware git:(12c1376) make
bundle exec rspec \
        --force-colour \
        --exclude-pattern 'spec/slow/**/*, spec/integration/**/*'

Randomized with seed 30547
..................................................................*...........................
```

</details>

spec/configurator_spec.rb
```diff
@@ -325,5 +329,36 @@ def configure_with_input(input_string)
...
+        configure_with_input("\n")
+      }.to raise_error(EOFError)
+    end
+
+    it 'allows optional questions to have empty answers' do
+      define_questions({
+        test: {
+          string_q: {
+            question: "I should NOT be re-asked",
+            optional: true
```
WilliamMcCumstie: updated the doc

src/configurator.rb
WilliamMcCumstie: Ahh `ActiveSupport` strikes again

src/configurator.rb
WilliamMcCumstie: updated

src/configurator.rb
WilliamMcCumstie: added

src/configurator.rb
WilliamMcCumstie: changed

WilliamMcCumstie:
Made those changes @bobwhitelock. Should be ready to merge.

spec/configurator_spec.rb
bobwhitelock: Thanks!

bobwhitelock:
Thanks for making all those changes, looks great!

PR #104, opened 2017-07-06
Fix build integration test failures

bobwhitelock:
@WilliamMcCumstie can you check that this fixes the issues for you on `master`?


WilliamMcCumstie:
LGTM, This fixes build failures in master. Will rebase the other hotfix to see if that fixes the other failures

Issue #102, opened 2017-07-04
metal ipmi command line weirdness


ste78:
[root@deploy ~]# metal ipmi -k 'sel list' db12mem04
Performing command on list..Address lookup for list.bmc failed
Could not open socket!
FAIL
Performing command on db12mem04..SEL Information
Version          : 1.5 (v1.5, v2 compliant)
Entries          : 15
Free Space       : 16144 bytes
Percent Used     : 1%
Last Add Time    : 07/04/2017 15:13:13
Last Del Time    : 06/01/2017 20:46:53
Overflow         : false
Supported Cmds   : 'Reserve'

Seems to have tried to run 'sel' on a server called 'list' ? IPMI commands can be quite long strings.

Also why have we dropped -c in favour of -k for the command argument?

BobWhitelock:
I'll leave this for @WilliamMcCumstie to investigate as he worked on it.

But regarding `-c` changing to `-k`: we have a `--config` option now which uses `-c` as a short option, presumably that's why its been changed. I didn't realise we had a `-c` option anywhere previously though; we should probably change it back as I don't think `--config` needs a short option.

ste78:
No need to change - that makes sense :)

WilliamMcCumstie:
Hmmm I think the problem might be when commander parses the `cli` it strips the quotes off the command. So the following command is being ran:
`/path-to/ipmi -k sel list db12mem04`

ColonelPanicks:
Issue testing output with @WilliamMcCumstie 
```
[root@deploy repo]# metal ipmi -k "'sel list'" db12mem04
Performing command on db12mem04..   1 | 06/01/2017 | 20:46:53 | Event Logging Disabled #0x72 | Log area reset/cleared | Asserted
   2 | 07/03/2017 | 19:27:43 | Fan #0x33 | Lower Non-critical going low  | Asserted
   3 | 07/03/2017 | 19:27:43 | Fan #0x33 | Lower Critical going low  | Asserted
   4 | 07/03/2017 | 19:28:06 | Fan #0x75 | Redundancy Lost | Asserted
   5 | 07/04/2017 | 12:29:16 | Fan #0x33 | Lower Non-critical going low  | Asserted
   6 | 07/04/2017 | 12:29:17 | Fan #0x33 | Lower Critical going low  | Asserted
   7 | 07/04/2017 | 12:29:42 | Fan #0x75 | Redundancy Lost | Asserted
[...]

PR #105, opened 2017-07-07
Fixes ipmi and a range of other bugs

WilliamMcCumstie:
The `repo_configs_path` config parameter has still be contributing to bugs in the tests despite being deprecated. This PR removes it completely from the code base.

Also the bug in `ipmi` was due to ruby stripping the quotes off the cli causing the command to be parsed as separate arguments in bash. This has now been fixed by directly passing the input array into `exec`.
Fixes: #102

### DIFF
--- a/spec/build_files_retriever_spec.rb
+++ b/spec/build_files_retriever_spec.rb
@@ -22,7 +22,8 @@
 
 require 'build_files_retriever'
 require 'input'
-
+require 'spec_utils'
+require 'config'
 
 RSpec.describe Metalware::BuildFilesRetriever do
   TEST_FILES_HASH = {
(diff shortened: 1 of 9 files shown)


bobwhitelock:
LGTM, thanks for tidying up the stuff with `repo_configs_path`.

I can't directly test the `ipmi` fix as I don't have any real nodes lying around, but I'm assuming this is still working fine for Stu and Steve.

bobwhitelock:
@WilliamMcCumstie: rebasing this on `develop` to merge it through there as well fails, due to conflicts in the templater spec - it looks like the underlying issue is you've defined two empty repos in different branches, `repo-empty` and `empty-repo`, and the same test is now trying to use both. Presumably this is not needed? Would you be able to do the rebase and standardise on one of these? Thanks.

PR #106, opened 2017-07-10
Improve filesystem testing

bobwhitelock:
Apologies for the massive PR, this ballooned a bit from what I originally intended :slightly_smiling_face:.

This key part of this PR is adding the `FileSystem` class, which similarly handles our interactions with `FakeFS` as the previous `FakeFSHelper` class did, but also does a bit more:

- ensures we are only using the `FakeFS` when necessary, when a `FileSystem.test` block is executing;

- provides more setup of the file system to be in the state we expect - the same files and directories that would be created by a normal Metalware install are now always created for `FileSystem` tests, as in practise we can reasonably expect these to always exist;

- provides more utility methods for setting up the `FakeFS` - for loading in sets of files from the `spec/fixtures` directory, for creating a minimal Metalware repo for testing purposes, and a `FileSystem.setup` method provides a way to setup the file system for groups of tests, while still using the real file system until tests execute;

- the `Debugging.print_fs` function has been improved and moved to `FileSystem.debug!`, for easier debugging of the current `FakeFS` state.

All uses of `FakeFSHelper` have been removed and replaced with this new class. Additionally, this PR also reduces some duplication in a few places, makes clearer when some tests use `FakeFS` and others in the same file do not (but perhaps should), and fixes a bug with checking dependencies on the repo (see https://github.com/alces-software/metalware/commit/4157bc28c876c5a2fcbfbb55279ce8cd202f3693).


spec/node_spec.rb
```diff
@@ -43,21 +38,86 @@ def node(name)
...
 
-  describe '#configs' do
-    it 'returns possible configs for node in precedence order' do
-      expect(testnode01.configs).to eq(["domain", "cluster", "nodes", "testnodes", "testnode01"])
+  before do
+    SpecUtils.use_mock_genders(self)
+  end
+
+  # XXX adapt these to use FakeFS and make dependencies explicit?
+  context 'without using FakeFS' do
```
WilliamMcCumstie: This maybe in a latter commit, but are their any tests that do use FakeFS?

spec/templater_spec.rb
```diff
@@ -71,7 +62,7 @@ def expect_renders(template_parameters, expected, config: Metalware::Config.new)
         alces.index: 0
         EOF
 
-        expect_renders({}, expected, config: @config)
+        expect_renders({}, expected)
```
WilliamMcCumstie: Where about are we starting FakeFS? I was expecting to see `FileSystem.test` somewhere

spec/filesystem.rb
```diff
@@ -75,6 +75,23 @@ def create_initial_directory_hierarchy
     FileUtils.touch Metalware::Constants::DEFAULT_CONFIG_PATH
   end
 
+  # Print every directory and file loaded in the FakeFS.
+  def debug!
+    begin
+      # This can fail oddly if nothing matches (see
+      # https://github.com/fakefs/fakefs/issues/371), hence the `rescue` with a
+      # simpler glob.
+      matches = Dir['**/*']
```
WilliamMcCumstie: If we want to print the entire directory structure, we can also use `Dir['/**/*']`. Not sure if that fixes the bug as we should always be able to find the root dir: `/`.

spec/templater_spec.rb
```diff
@@ -24,29 +24,31 @@
...
 
 
 RSpec.describe Metalware::Templater do
-  def expect_renders(template_parameters, expected, config: Metalware::Config.new)
-    fshelper = FakeFSHelper.new(config)
-    fshelper.clone(FIXTURES_PATH)
-    fshelper.clone_repo(config.repo_path)
+  let :filesystem {
+    FileSystem.setup do |fs|
+      fs.with_fixtures('/', at: '/fixtures')
```
WilliamMcCumstie: I am finding it a tad confusing what you are trying to achieve here. My understanding is that `fs.with_fixtures` clones files from the real `fixtures` dir to the location specified by at. However methods cached by `FileSystem.setup` run after the `FakeFS` is in place and thus you no longer have access to the real files.

spec/dependencies_spec.rb
```diff
@@ -24,22 +24,24 @@
...
       Metalware::Dependencies.new(config, "test", dependencies_hash).enforce
     end
   end
 
-  context 'with a blank filesystem' do
-    it 'repo dependencies fail' do
+  context 'with a fresh filesystem' do
+    # XXX Skipped as, with filesystem after fresh install, repo dependencies
+    # specified with an empty array should not pass, but they will.
+    xit 'repo dependencies fail' do
```
WilliamMcCumstie: Why would this cause an issue in the new `FileSystem` testing? `repo` dependencies should be checking the existence of a directory

spec/dependencies_spec.rb
```diff
@@ -39,9 +39,7 @@ def enforce_dependencies(dependencies_hash = {})
   end
 
   context 'with a fresh filesystem' do
-    # XXX Skipped as, with filesystem after fresh install, repo dependencies
-    # specified with an empty array should not pass, but they will.
-    xit 'repo dependencies fail' do
+    it 'repo dependencies fail' do
       expect {
```
WilliamMcCumstie: Ignore previous comment

spec/node_spec.rb
bobwhitelock: Yep, in the final PR the `Node#answers` test do. I've been including these comments when I spot some tests which depend on the file system but do not use `FakeFS`, and instead use some other means of working with paths, to remind us to fix this if these are touched again.

spec/templater_spec.rb
bobwhitelock: `FileSystem#with_fixture` is a wrapper around `FakeFS::FileSystem.clone`, which can run whether or not the `FakeFS` is activated and will clone files into it; I believe it bypasses `FakeFS`'s own replacements of various methods for interacting with the file system. So this works fine whether you run it in `setup`, `test`, or on a `FileSystemConfigurator directly. You can see this if you comment this line, then the tests will fail with various errors like this:

```
     Errno::ENOENT:
       No such file or directory - /fixtures/template.erb
```

spec/templater_spec.rb
bobwhitelock: If you look in `expect_renders` you'll see `filesystem.test`; this runs `FileSystem.test` but using the `FileSystemConfigurator` built up in in the `let :filesystem` block above. See https://github.com/alces-software/metalware/blob/175a29421a732445feaaf14adbf1f9b2896d32a9/spec/filesystem.rb#L154-L156 for its source.

spec/filesystem.rb
bobwhitelock: Thanks for the suggestion - this doesn't appear to fix it however; if you add `p Dir['/**/*']` to the first `it` in the issue I opened (https://github.com/fakefs/fakefs/issues/371) it will still raise an error.

I think this is related to the way shell-globbing works, which `Dir.[]` attempts to match, e.g. if you run the following from the Metalware repo root:
```
$ echo src/commands/repo/**/*
src/commands/repo/update.rb src/commands/repo/use.rb
```
- you can see that the directory `src/commands/repo/` itself is not matched by the glob.

bobwhitelock:
Thanks for the review @WilliamMcCumstie!

Regarding your comment about `FileSystem.setup`: essentially I wanted to create an API which allows us to run a test in an environment like a fresh Metalware install with just `FileSystem.test do ...`, but we also sometimes need to perform setup shared between several tests before doing this. To do this I added `FileSystem.setup`, which returns a `FileSystemConfigurator` object. This saves up any calls made to it, and also has a `test` method of its own.

When `test` is called on a `FileSystemConfigurator` object, it calls `FileSystem.test` and passes itself, so `FileSystem.test` will then run and invoke the saved methods before `yield`ing to the block passed to `test`. When `FileSystem.test` is called directly a new `FileSystemConfigurator` instance is used instead, which has no saved methods and so no additional setup is performed.

So to answer your question directly: each call to `FileSystem.setup` creates a new `FileSystemConfigurator` object. This will also be yielded to a block if one is passed to `FileSystem.setup`, but this is just syntactic sugar over needing to go `FileSystem.setup.tap do |filesystem_configurator| ...` (or `FileSystemConfigurator.new.tap do |filesystem_configurator| ...`) all the time, as we often want to call setup methods immediately when setting up the file system.

We could change the API to have you just create `FileSystemConfigurator` instances directly and pass them to `FileSystem.test` - I was trying to avoid this though as I thought it was nicer to not expose `FileSystemConfigurator`, and you should usually only need to be aware of the `FileSystem` class.

Do you think we need any changes before merging this?

spec/templater_spec.rb
WilliamMcCumstie: Looks good, it was in a latter commit

WilliamMcCumstie:
Will need to see the `FileSystemConfigurator` in action, but as all the tests are passing it looks good to me.

PR #107, opened 2017-07-10
Improve magic namespace tests

bobwhitelock:
Based on https://github.com/alces-software/metalware/pull/106. This is just some refactoring of the Templater's magic namespace tests, to in particular use `FakeFS` and `FileSystem`. Submitting this now to keep it separate from my next PR which will also touch the `MagicNamespace`.

### DIFF
--- a/spec/templater_spec.rb
+++ b/spec/templater_spec.rb
@@ -164,11 +164,13 @@ RSpec.describe Metalware::Templater do
       expect(genders_config.non_existent).to eq([])
     end
 
-    before do
-      # Stub this so mock hunter config used.
-      stub_const('Metalware::Constants::HUNTER_PATH', TEST_HUNTER_PATH)
+    let :filesystem {
+      FileSystem.setup do |fs|
+        fs.with_repo_fixtures 'repo'
+      end
+    }
 
-      SpecUtils.use_unit_test_config(self)
+    before do
       SpecUtils.use_mock_determine_hostip_script(self)
       SpecUtils.use_mock_genders(self)
     end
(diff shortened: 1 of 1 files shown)


WilliamMcCumstie:
Rebased on develop

spec/templater_spec.rb
```diff
@@ -164,11 +164,13 @@ def expect_environment_dependent_parameters_present(magic_namespace)
...
+    let :filesystem {
+      FileSystem.setup do |fs|
+        fs.with_repo_fixtures 'repo'
+      end
+    }
 
-      SpecUtils.use_unit_test_config(self)
+    before do
       SpecUtils.use_mock_determine_hostip_script(self)
       SpecUtils.use_mock_genders(self)
```
WilliamMcCumstie: This is probably a future change, but do we need to mock out the `genders` and the `hostip_scripts` anymore if we are using FakeFS? It would be cleaner if we could integrate them with `FileSystem` if appropriate.

WilliamMcCumstie:
LGTM

spec/templater_spec.rb
bobwhitelock: I agree it would be nice to handle these in tests a better way, but I don't think `FakeFS` helps us here.

For the `determine-hostip` script, it doesn't look like `FakeFS` changes Ruby functions for running external commands to use the `FakeFS` - this is probably good in general, I imagine there would be a lot of odd cases and surprising behaviour if it did.

And then for using `genders`, we never actually load the `genders` file directly ourselves, rather we rely on `nodeattr` to parse it and query that for the results. Even if we loaded the `genders` file into the `FakeFS`, there would be no way to tell `nodeattr` to use this - the `FakeFS` isn't even a real file system; you'd need to have `nodeattr`, a non-Ruby program, somehow load the `genders` file from Ruby's memory!

There may be some improved abstraction we can use to paper over this sort of thing, but I agree we can do that in future as needed.

PR #111, opened 2017-07-11
Validates configure.yaml string questions

WilliamMcCumstie:
This is the first stage to validating the `configure.yaml` questions. This PR is based on #103 which needs to be merged first.

The two major changes to the code is to allow `Dependency` to run a code block when validating if a file exists. (NOTE: `Dependencies` has been renamed to `Dependency` so it is singular). This allows specific file content to be validated using arbitrary code.

The second change is the introduction of a `Validator::Configure` class. This exists in a new `Validator` module which can be expanded to support other validations. Their is a `BaseValidator` class which can be used to store code that can be used by all future `Validators`. The basic design is that `valid?` method can be called on a validator which returns true or false.

In `Configure` their is a `validate` method that runs a set of tests on the input. Currently it only confirms that `domain`, `group` and `node` exist as top level keys (`question` is optional) and that string type questions have string defaults.

All other types will currently causes the validator to fail and hence it was necessary to mock out the validator is the `each`, `build`(, etc) tests. Future PR's will introduce validation for other question types.

NOTE: This PR merges into `feature/validate-configure` which is currently up to date with develop. Once all the data type PR's are merged into this branch, it can be merged into `develop`

### DIFF
--- a/spec/commands/build_spec.rb
+++ b/spec/commands/build_spec.rb
@@ -73,7 +73,7 @@ RSpec.describe Metalware::Commands::Build do
     SpecUtils.use_mock_genders(self)
     SpecUtils.use_unit_test_config(self)
     SpecUtils.fake_download_error(self)
-    SpecUtils.use_mock_dependencies(self)
+    SpecUtils.use_mock_dependency(self)
   end
 
   context 'when called without group argument' do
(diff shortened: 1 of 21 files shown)


WilliamMcCumstie:
The validator class is going to be replaced with `Dry::Validation` which renders this PR obsolete

PR #110, opened 2017-07-10
Add `group_index`

bobwhitelock:
Based on https://github.com/alces-software/metalware/pull/107.

This PR adds a `group_index` parameter to the `alces` templater namespace, which returns a unique, consistent, index for the current node's primary group. This forms another part of the work towards https://trello.com/c/XjtdbEHm.

PR #112, opened 2017-07-11
Improve and standardise YAML handling

bobwhitelock:
Based on branch under review in https://github.com/alces-software/metalware/pull/110.

This PR improves and standardises how we handle the various Metalware YAML data files, using a new `Data` module. Pretty much all our current and future working with YAML should go through this module, which is very simple but means we always do things in a standard way and do not need various ad hoc handling and conversion throughout Metalware.

Using this module will recursively convert all hash keys to symbols on `load` and recursively convert hash keys to strings on `dump` (for consistency and slightly easier user editing); `load` also returns an empty hash when the data file does not exist or is empty.


bobwhitelock:
(rebased on `develop` to make this easier to review)

bobwhitelock:
(rebased again to fix failing test after the previous rebase :stuck_out_tongue:)

WilliamMcCumstie:
I am still getting failing tests:
```
Failed examples:

rspec ./spec/data_spec.rb:57 # Metalware::Data#dump dumps the data to the data file with all keys as strings
rspec ./spec/data_spec.rb:47 # Metalware::Data#load raises if the file contains invalid YAML
rspec ./spec/data_spec.rb:33 # Metalware::Data#load returns {} if the file is empty
rspec ./spec/data_spec.rb:25 # Metalware::Data#load loads the data file and recursively converts all keys to symbols
```

WilliamMcCumstie:
Looks like you have a hard coded path or haven't faked the `fs` correctly:
```
1) Metalware::Data#dump dumps the data to the data file with all keys as strings
     Failure/Error: Metalware::Data.dump(data_file_path, symbol_keyed_data)
     
     Errno::ENOENT:
       No such file or directory - /home/some_data.yaml
```

bobwhitelock:
@WilliamMcCumstie: weird - have you `pull`ed this recently? These pass on my machine.

WilliamMcCumstie:
Just now

bobwhitelock:
Ah, I think I see what the problem might be - `FakeFS` implicitly creates the path to the current directory in the fake file system. When I test on my machine my code is within `/home/bob/...`, but on the Controller it's in `/tmp/metalware`, so `/home` isn't created and it can't clone a file at that path :disappointed:. Let me just fix it.

bobwhitelock:
I've added another commit which should fix that, can you check if it does for you? Also fair enough about `CliParser`, I'll add that shortly.

WilliamMcCumstie:
Ooops I jumped the gun a bit. I'll add the comment in my branch, I need to rebase on this branch anyway

PR #113, opened 2017-07-12
Extract Templater helpers

bobwhitelock:
Based on https://github.com/alces-software/metalware/pull/112.

This PR extracts the various `Templater` helpers from being nested within it to a new `Templating` module. This was prompted by needing to use the `MagicNamespace` from elsewhere (as part of adding rendering of `genders`, and `hosts` in one pass), and so it made sense to extract it rather than needing to namespace it by the `Templater`. In the process it then made sense to extract the other `Templater` helper classes, as these and the `MagicNamespace` reference each other.

This resolves https://trello.com/c/H2Fjif5x/113-extract-magicnamespace-from-templater.


src/templater.rb
```diff
@@ -19,16 +19,16 @@
...
-require "ostruct"
-require "yaml"
 
 require "constants"
 require 'metal_log'
 require 'exceptions'
 require 'iterable_recursive_open_struct'
 require 'magic_namespace'
 
+
```
WilliamMcCumstie: Do we need the space?

WilliamMcCumstie:
LGTM, just need to merge #112 first

WilliamMcCumstie:
Rebased on develop. Otherwise LGTM

src/templater.rb
bobwhitelock: Not really, I just like to leave a couple of spaces between my `require`s and the start of a class definition! Probably a habit picked up from [Python](https://www.python.org/dev/peps/pep-0008/#blank-lines).

PR #114, opened 2017-07-12
Validate the configure.yaml files

WilliamMcCumstie:
Based #112, please merge that first.

This PR contains most of the validation required for the `configure.yaml` files. In the end, the layout of the `configure.yaml` has not been altered, but this still an option going into the future.

Setting up to allow this change, the `Dependencies` class needed to be altered. It is now possible to run a block of code to validate a file. `Dependencies` has also been renamed to be `Dependency`.

Following this, the `configure.yaml` file is validated using `Dry::Validation`. Please refer to individual commits on how this is completed. At this stage `choice` questions are not completely validated as this will occur in a separate commit.

However default values are validate against the question type. Note that only `integer` and `string` defaults are currently excepted.


WilliamMcCumstie:
Rebased on `dev/improve-yaml-handling` so that the hash can be converted to `symbols`

bobwhitelock:
Rebased on `develop` to resolve a lot of conflicts.

bobwhitelock:
@WilliamMcCumstie: this looks good to me, although I've not looked in depth at the dry-validation logic - it looks pretty crazy :slightly_smiling_face:. I've added a few commits to fix things after rebasing on my recent changes, as well as a couple of other minor tweaks.

A couple of other general comments:

- Possibly a better name for the `Dependency` class would be `DependenciesValidator`, as it is not really a 'dependency' itself but it does validate them - what do you think?

- In what situation do the error messages defined in `src/validator/errors.yaml` get used? I tried adding a bad key to a question and then ran `configure domain`, but the only output I got was:

  ```
  error: 'configure domain' requires a valid 'repo/configure.yaml' file. Use --trace to view backtrace
  ```
  Not sure if this is supposed to be the error message we get in this case, but either way how hard would it be to change this to give more indication of what makes the file invalid? Currently it does not give much to go on.

Going to merge this through now, but let me know about the above when you're back.

PR #115, opened 2017-07-12
Add `groups` to `alces` namespace

bobwhitelock:
Based on https://github.com/alces-software/metalware/pull/113.

This PR adds a `groups` function to the `alces` namespace, as discussed on Slack at https://alces.slack.com/archives/C5FL99R89/p1499856806970388. Currently `groups` will call the passed block once for each primary group (each group already configured with `metal group configure`), with each call passed a `GroupNamespace` for the current group. The `GroupNamespace` currently provides two properties: `name`, the group name, and `answers`, the `configure` answers for that group; this is the domain answers with the group answers merged in, as these are the only answers which are applicable to a particular primary group.

Access to the `nodes` will also be needed from a `GroupNamespace`, but this will follow in a future PR as this is already quite large; apart from this I don't think we will need anything else in this namespace for the moment in order to render the `genders` or `hosts` templates.

In the process of doing this I found there were a few things on `Node` which needed to be extracted, as we needed to use the same/similar functionality in other places. This resulted in extracting a `Templating::Configuration` class, which I think seems an OK abstraction, as well as a `PrimaryGroup` class, which I think will need a bit more work. I think there's probably a better way we could divide up the responsibilities of `Nodes`, `Templating::GroupNamespace`, and `PrimaryGroup`, but I'm not quite sure what that is yet so I'm leaving it for now.

This is part of the way towards handling https://trello.com/c/cj9s16dR/77-consider-how-to-make-groups-and-nodes-available-in-templater.

### DIFF
--- a/spec/fixtures/answers/group_namespace_tests/domain.yaml
+++ b/spec/fixtures/answers/group_namespace_tests/domain.yaml
@@ -1,0 +1,3 @@
+
+domain_value: 'domain_value'
+overriding_domain_value: 'domain_value'
--- a/spec/fixtures/answers/group_namespace_tests/groups/testgroup.yaml
+++ b/spec/fixtures/answers/group_namespace_tests/groups/testgroup.yaml
@@ -1,0 +1,3 @@
+
+overriding_domain_value: 'testgroup_value'
+genders_host_range: 'node0[10-20]'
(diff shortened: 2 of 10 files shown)


WilliamMcCumstie:
This branch has some conflicts likely due to the rebasing. Will probably need to be rebased again on develop. #113 has been merged

bobwhitelock:
@WilliamMcCumstie: rebased :slightly_smiling_face:

WilliamMcCumstie:
LGTM

PR #116, opened 2017-07-13
Extract more templater parts

bobwhitelock:
Based on https://github.com/alces-software/metalware/pull/115.

This PR extracts some more parts of the `Templater` to two new classes, `RepoConfigParser` and `Renderer`, along with some other small refactoring. This is done as we need to use this functionality in other places now, and it makes sense to split this up a bit more in any case.

### DIFF
--- a/src/templater.rb
+++ b/src/templater.rb
@@ -61,28 +61,17 @@ module Metalware
     end
 
     attr_reader :config
-    attr_reader :nodename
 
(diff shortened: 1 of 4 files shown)


WilliamMcCumstie:
Needs to be rebased

bobwhitelock:
Rebased; going to merge this for now but feel free to have you a look through when you're back if you like :slightly_smiling_face:.

PR #128, opened 2017-07-25
Add command to view templating config

bobwhitelock:
This PR adds a new command, `metal view-config` which shows the configured Metalware templating config, optionally in the context for a particular node. This shows as JSON the exact config which will be used when rendering templates in general (as used in `dhcp`, `hosts`, and `render` without a node passed) or for a particular node (as used in `build` and `render` with a node passed).

This replaces and supersedes the previous `show_config` script I threw together before, with a few improvements:

- this is an actual Metalware command, so it is always available rather than needing to copy scripts around;

- the full templating config is now shown, including the `alces` namespace and optionally the `alces.groups` namespaces and nodes within these;

- syntax highlighting of the output JSON is provided via [jq](https://stedolan.github.io/jq/) - additionally since `jq` is now installed along with Metalware, this also means you can pipe the output of this command to it as needed to further filter the output to get the parts of the config you want, e.g. the following command just shows the `alces` namespace:

```
   [root@controller(dev) metalware]# bin/metal view-config node01 | jq .alces
{
  "answers": {
    "cluster_name": "foo",
    "domain": "<%= alces.answers.cluster_name %>.cluster.local",
    "root_password": "password from domain",
[ ... etc ... ]
```

See the [description](https://github.com/alces-software/metalware/compare/develop...feature/viewing-templating-config#diff-3d8becb4f43e68f36881f3f383d363a7R307) for more details of what the command shows and some limitations.

@ColonelPanicks: I've added you as a reviewer in case you want to have a look before this is merged, and see how this fulfils your use cases for checking the config when working with Metalware. If you do you may want to install Metalware from the next PR instead, as that will be based on this one but with additional commands for viewing the answers.

PR #129, opened 2017-07-25
Add command to view configure answers

bobwhitelock:
This PR is based on and builds upon https://github.com/alces-software/metalware/pull/128. It adds several sub-commands under a new command, `view-answers`, to view the configured answers for a particular object in Metalware, as well as the answers at higher levels which will also apply to this object when it has unset answers.

This command was requested in the meeting last week; I thought it was useful to have both these and the `view-config` command, as that allows you to view the exact complete templating config, which is useful for debugging, while these commands let you to get a better overview of what answers will be used. This is more useful for normal Metalware usage as that is all the typical user should need to configure to build a cluster.

@ColonelPanicks: similarly to the previous PR, I've added you as a reviewer if you want to try out how this works in practise.


src/config.rb
```diff
@@ -70,6 +72,18 @@ def configure_file
       File.join(repo_path, 'configure.yaml')
     end
 
-    attr_reader :cli
+    def domain_answers_file
```
WilliamMcCumstie: In my answer validator, I need to know the relative paths for each of the `Configure` commands inside the answer `directory`. So I have already added a relative path function where required. As a result we don't need three separate method to handle each of the sections. One methods should be sufficient.

src/node.rb
```diff
@@ -68,8 +73,14 @@ def built?
...
-      # not part of any groups.
+      # Re-raise if we're expecting the node to be configured, as it should be
+      # in the genders file.
+      raise if should_be_configured
+
+      # The node doesn't need to be in the genders file if it's not expected to
+      # be configured, it's just part of no groups yet (XXX not sure if this is
+      # still true, maybe we should be stricter now and always require a node
+      # to be in the genders file).
       []
```
WilliamMcCumstie: It's probably safer if we always require the node to be in the genders file. Otherwise it would imply that we have a node without a primary group.

src/repo.rb
```diff
@@ -0,0 +1,26 @@
+
+# frozen_string_literal: true
+
```
WilliamMcCumstie: Needs a copyright notice

src/templating/group_namespace.rb
```diff
@@ -6,6 +6,16 @@ module Templating
     class GroupNamespace
```
WilliamMcCumstie: copyright notice

src/templating/magic_namespace.rb
```diff
@@ -14,15 +14,52 @@
 module Metalware
   module Templating
     class MagicNamespace
-      def initialize(config:, node: nil, firstboot: nil, files: nil)
+      # Every time a new property is added to this namespace it should be added
+      # to this array (so it shows in the `view-config` output), as well as to
+      # the docs in `docs/templating-system.md`.
+      FIELDS = [
```
WilliamMcCumstie: Is their a more implicit way to do this? What is the criteria to be included? Could we make it all public methods that are not part of `Object.methods`?

src/commands/configure/domain.rb
```diff
@@ -13,7 +13,7 @@ def setup(_args, _options); end
         protected
 
         def answers_file
-          File.join(config.answer_files_path, 'domain.yaml')
+          config.domain_answers_file
```
bobwhitelock: Not sure what you mean - it's domain-specific because it returns the domain answers file and not any other. I think there's an advantage to having a method which takes no arguments and returns the correct file, rather than a method we need to ensure we pass the correct relative path to to get the domain answers file. Of course any of these answer file methods can use the same method to get the full path behind the scenes.

src/config.rb
bobwhitelock: As above - I think it's clearer/ less likely to lead to bugs to have methods to get each of the different answer paths, but we can expose a shared method used by each of these for the validator if you like.

src/commands/configure/domain.rb
WilliamMcCumstie: Ok so leave it as is. But I may need to change it when it comes to merging #134. This abstracts away some of the information I need.

src/node.rb
bobwhitelock: I think you may be right, I've added a card to track this here: https://trello.com/c/mMvRUUxw/149-always-require-created-nodes-exist-in-the-genders-file

src/templating/magic_namespace.rb
bobwhitelock: Hmm, this is a nice idea; let me investigate.

bobwhitelock:
@WilliamMcCumstie: I've added a PR to add what you suggested above (made sense to do this separately as the code you commented on was actually in the previous PR); since you're now adding the copyright notices is this mergeable?

WilliamMcCumstie:
So the `rubocop` copyright feature doesn't seem to be working for me. I am just going to use my previous technique of adding them when I come across the files.

PR #130, opened 2017-07-26
Disallow configuring groupless node

bobwhitelock:
Based on https://github.com/alces-software/metalware/pull/129; this resolves https://trello.com/c/NB83ojWQ/130-it-should-not-be-possible-to-configure-a-node-which-isnt-in-a-group.

This PR resolves the above card by not allowing configuring a node which is not part of an already configured group. This is done by extracting the existing `dependency_hash` generating where we do this from `Commands::ViewAnswers::Node` to a new class, and then using this both here and in `Commands::Configure::Node`.


spec/dependency_specifications_spec.rb
```diff
@@ -0,0 +1,30 @@
+
```
WilliamMcCumstie: Copyright notice

spec/repo_spec.rb
```diff
@@ -0,0 +1,32 @@
+
```
WilliamMcCumstie: copyright notice

src/dependency_specifications.rb
```diff
@@ -0,0 +1,31 @@
+
```
WilliamMcCumstie: Copyright notice

src/dependency_specifications.rb
```diff
@@ -0,0 +1,31 @@
...
+  class DependencySpecifications
+    def initialize(config)
+      @config = config
+    end
+
+    def for_node_in_configured_group(node_name)
+      primary_group = configured_primary_group_for_node(node_name)
+      {
+        repo: ['configure.yaml'],
+        configure: ['domain.yaml', "groups/#{primary_group}.yaml"],
```
WilliamMcCumstie: I think this also needs `nodes/#{node}.yaml` to insure that the node answers are validated. We could possible use this for validating the configure commands.

It might be also useful to include a group option

bobwhitelock: Don't think we want to include `nodes/#{node}.yaml` - every node needs to be part of a configured group, so we need `groups/#{primary_group}.yaml`, but each node in a group does not need to have `metal configure node` run for it, so most `nodes/#{node}.yaml` files will not be found and the dependencies will fail when they should have passed.

WilliamMcCumstie: Ok, but we will need to validate the files if they exist, so that we can insure that the types are correct.

Maybe we could include an `optional` hash. Where any files included do not have to exist, but if they do, additional validation is required. Their are a few locations where we need to do something along these lines.

bobwhitelock: OK - adding a distinct method for that sounds reasonable, otherwise it's going to make things quite a bit more complicated if we start having a lot of optionally added files in the dependencies.

WilliamMcCumstie: I was thinking changing the dependency hash something like this:
```
{
  repo: ['configure.yaml'],
  configure: ['domain.yaml', ... ],
  optional: {
    configure: ['nodes/#{node}.yaml']
  }
}
```

Then we can have `Dependency` work it all out for us

bobwhitelock: @WilliamMcCumstie: Yep, that format looks reasonable :slightly_smiling_face:

WilliamMcCumstie: Ok once #138 is merged we can fix this and merge

bobwhitelock: I tried to make this change but I get the following error when I do so and then run a command using the dependencies:
```
[root@controller(dev) metalware]# bin/metal view-answers node node02
error: Dependency values must be an array, check: optional. Use --trace to view backtrace
```

Since this is the only thing blocking this I'm going to merge it as is, but when you've looked at that error we can add in that change; it's in https://github.com/alces-software/metalware/tree/wip-adding-optional-dependencies. Cards for this: https://trello.com/c/fnNhvA22/153-fix-specifying-optional-dependencies (for you to look at) and https://trello.com/c/2tL6Q2CP/154-add-in-optional-dependency-for-nodes-on-node-answers (to add my change in once that's done).

WilliamMcCumstie: Looks like that branch is missing the following commit. Should work otherwise.
https://github.com/alces-software/metalware/pull/138/commits/717207279fde9dad94725304eec7e410b3ade831

PR #132, opened 2017-07-26
Add managed file message to domain files

bobwhitelock:
Based on #131; this resolves https://trello.com/c/tCUxtbwu/132-add-note-to-rendered-hosts-and-genders-or-possibly-all-rendered-templates-that-template-was-rendered-from-metalware-and-changes.

PR #133, opened 2017-07-26
Remove hosts

bobwhitelock:
Based on #132.

With the addition of the `hosts` file being rendered in full as `configure` commands are run, the `hosts` command is now obsolete, and in fact always errored currently due to the changed template. This commit removes it, its tests, and the `Templater.render_and_append_to_file` method only used by it.


bobwhitelock:
Rebased for easier reviewing

PR #134, opened 2017-07-27
Validate answers file

WilliamMcCumstie:
This PR contains the code for validating the answer files. Due to the complexity of the hash, this had to be done in multiple steps. For consistency, all validation is done using Dry::Validation so the return object will always behave in a similar way. 

The first step uses `AnswerBasicSchema` to confirm that the answers is a `Hash`. This is required as the answer hash is then merged with the quetion hash from `configure.yaml`. The combined hash is used as it is in a format that is easier to validate. 

Next `MissingSchema` is used to detect answers to missing questions. These answers are set aside as part of the combining process which is then detected by the schema.

The last stage is when `AnswerTypeSchema` loops through the questions and confirms their types.

A few pedantry changes where required to the Makefile and .gitignore to make rubocop work locally on my machine. Also the Configure commands now require a valid answer file if it exists as it is used to determine the defaults.


WilliamMcCumstie:
This will need to be rebased and integrated with #138

PR #135, opened 2017-07-27
Dynamically decide namespace fields to hash

bobwhitelock:
This commit introduces a module with a single function which will take an object and convert its methods and their values to the keys and values of a hash, with the optional ability to override what method is
used to get certain fields. This function is then used in the `MagicNamespace` and `GroupNamespace` when we convert them to hashes, so we don't need to manually specify the fields to include.

This function will of course only work for objects where all methods to be included can take no arguments, otherwise it will blow up as the methods won't be sent enough arguments.

As suggested in https://github.com/alces-software/metalware/pull/129#discussion_r129885647.

### DIFF
--- a/spec/object_fields_hasher_spec.rb
+++ b/spec/object_fields_hasher_spec.rb
@@ -1,0 +1,37 @@
+
+# frozen_string_literal: true
+
+require 'object_fields_hasher'
+
+RSpec.describe Metalware::ObjectFieldsHasher do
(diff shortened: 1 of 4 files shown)


WilliamMcCumstie:
LGTM, in the case where a function takes more than one input, we could use the override somehow to get the value. However this is probably good for now. We can cross that bridge when we get to it

PR #138, opened 2017-07-28
Allow optional files in Dependency

WilliamMcCumstie:
A new optional hash can be included in the dependency_hash which indicates files that should be validated but does not throw an error if they do not exist.

The optional hash has the same syntax as a regular dependency_hash. At this point in time, the superfluous, however it will be needed once Answer Validation is included.


bobwhitelock:
LGTM

PR #139, opened 2017-07-28
Obey hierarchy when configuring

bobwhitelock:
Based on https://github.com/alces-software/metalware/pull/133. This PR resolves https://trello.com/c/75bjGbhK and https://trello.com/c/jV307Gji.

When configuring at each level, we now only save an answer if an answer is entered and this is different to the answer at a higher level in the answer files hierarchy - note that the answer files hierarchy currently goes node > group > domain > question default, from highest precedence to lowest. In order to have this work when templating we also now need to load the question defaults when forming the overall answers, since these are no longer being saved in the answer files.


src/commands/configure/node.rb
```diff
@@ -18,10 +18,21 @@ def answers_file
...
         private
 
         attr_reader :node_name
 
+        def node
+          Metalware::Node.new(config, node_name)
+        end
+
         def dependency_hash
           dependency_specifications.for_node_in_configured_group(node_name)
```
WilliamMcCumstie: This isn't going to generate the correct `dependency_hash` anymore. Now that that `optional` dependencies are supported, this method skips the checking of `nodes/[node_name].yaml`

WilliamMcCumstie: Also I don't think `dependency_specifications` is even required anymore as it can be replaced with the `higher_level_answer_files`

src/commands/configure/node.rb
```diff
@@ -18,10 +18,21 @@ def answers_file
           config.node_answers_file(node_name)
         end
 
+        def higher_level_answer_files
+          [
+            config.domain_answers_file,
+            config.group_answers_file(node.primary_group)
+          ]
+        end
```
WilliamMcCumstie: These need to be included as an optional dependency. It can probably be placed in `ConfigureCommand` so the inherited commands do not need to worry about it

src/command_helpers/configure_command.rb
```diff
@@ -29,6 +29,10 @@ def answers_file
         raise NotImplementedError
       end
 
+      def higher_level_answer_files
+        []
+      end
+
       private
```
WilliamMcCumstie: Add `higher_level_answer_files` to the dependency hash. Something like:
```
def dependency_hash
  repo: ['configure.yaml']
  optional: {
    configure: ['domain.yaml', relative_answer_file].join(higher_level_answer_files)
  }
end

WilliamMcCumstie: Note: #134 will need to be merged and rebased

PR #144, opened 2017-08-01
Run every test using FakeFS unless specify not to

bobwhitelock:
Run every test using `FakeFS` unless a truthy `real_fs` metadata value is passed; this prevents us polluting the real file system unless explicitly requested.

With this change it may be possible to simplify our use of `FileSystem` to not need to re-specify running in a `FakeFS` block around every test. However this would be quite a large refactoring and I'm not sure exactly how we'd want to do this at the moment, so I'm holding off on this until we've used this change for a bit.

### DIFF
--- a/spec/build_files_retriever_spec.rb
+++ b/spec/build_files_retriever_spec.rb
@@ -27,7 +27,7 @@ require 'input'
 require 'spec_utils'
 require 'config'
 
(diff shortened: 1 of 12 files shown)


WilliamMcCumstie:
LGTM

PR #146, opened 2017-08-02
Created a unified Loader class for validation files

WilliamMcCumstie:
Due to the extensive use of Validation on files, the `Dependency` class is becoming unyielding. An alternative to this is to use a unified `ValidatorLoader` that simplifies how a file can be loaded and auto-magically validates the file.

The loader has been built so that it can incorporate `save` methods as well as it logical for these to be located with the `load` methods. The `load` and `save` methods are handled by different classes as the steps involved are different and hard to generalise due to validation. However they both inherit from a shared base class that contains the path information.

The syntax for using the `Loader` is as follows:
```
ValidatorLoader.<load|save(data)>.<file_name_method(*additional_parameters)>
```

The file loader can be created from the `config` and exists natively inside the `BaseCommand`.

Also updated the `Commands` class to it dynamically requires all `commands` in the `src/commads` directory. This should make it easier to add and remove commands in the future.


WilliamMcCumstie:
NOTE: Codeclimate is failing due to the `TODO's` that are in `Validator::Answer`. However fixing all the required changes is to big of a task for a single PR in will be fixed in the future

src/validator/answer.rb
```diff
@@ -30,9 +30,10 @@ module Validator
     class Answer
       ERROR_FILE = File.join(File.dirname(__FILE__), 'errors.yaml').freeze
 
-      def initialize(metalware_config, answer_file)
+      def initialize(metalware_config, answer_file, loader: nil)
```
WilliamMcCumstie: This was added before the `Loader` was in the `config`. We clean that up when I clean the class up as a whole

bobwhitelock: OK, sorry I must have missed that :slightly_smiling_face:

WilliamMcCumstie:
Ok I have done a bit more work on the `Loader` and `Saver` and simplified the syntax. The multiple class have been separated into individual files and the purpose has been refined. I have removed a lot of the meta-programming to make it easier. 

I have now created a `FilePath` class to simplify getting the paths to all the files. It is for the most part and extension of `<section>_answer_file` methods in `Config`. However I feel that these methods would be better served in a dedicated class. `FilePath` does have some meta-programming in which it can access the paths stored in `Constants`. It does make the assumption that any constant ending in `_PATH` is indeed a path.

I have also made all the `TODO` changes to `Answers` and reflected the change in `Dependency`

WilliamMcCumstie:
tl;dr; Everything has changed and you will need to review it again

src/validator/saver.rb
```diff
@@ -0,0 +1,55 @@
...
+# For more information on the Alces Metalware, please visit:
+# https://github.com/alces-software/metalware
+#==============================================================================
+
+require 'file_path'
+require 'data'
+
+module Metalware
+  module Validator
+    class Saver
```
bobwhitelock: Do we use this class anywhere (outside the tests for it)? If not can we remove.

src/validator/loader.rb
```diff
@@ -0,0 +1,70 @@
...
+# https://github.com/alces-software/metalware
+#==============================================================================
+
+require 'file_path'
+require 'validator/answer'
+require 'validator/configure'
+require 'data'
+
+module Metalware
+  module Validator
```
bobwhitelock: As mentioned in previous review, I think `Validation` might be a better name for this module.

src/validator/loader.rb
```diff
@@ -0,0 +1,70 @@
...
+    class Loader
+      def initialize(metalware_config)
+        @config = metalware_config
+        @path = config.file_path
+      end
+
+      def configure
```
bobwhitelock: `configure_data` might be a better name, calling this `configure` implies to me that this will do something like configure the `Loader` instance. You could suffix all these methods with `_data` if you want to be consistent.

src/validator/loader.rb
```diff
@@ -0,0 +1,70 @@
...
+        Validator::Configure.new(path.configure_file).load
+      end
+
+      def groups_cache
+        Data.load(path.groups_cache)
```
bobwhitelock: Don't think this method needs to exist, we don't use it outside of tests (afaics) and its existence on the `Validator::Loader` class implies it will validate the loaded file, which it doesn't do.

src/file_path.rb
```diff
@@ -0,0 +1,66 @@
...
+      config.domain_answers_file
+    end
+
+    def group_answers(group)
+      config.group_answers_file(group)
+    end
+
+    def node_answers(node)
+      config.node_answers_file(node)
+    end
```
bobwhitelock: If you give these methods the same name here and in the `Config` class, then you can simplify all these methods to just `delegate` to the `config`.

src/command_helpers/base_command.rb
```diff
@@ -75,6 +75,10 @@ def enforce_dependency
         Dependency.new(config, command_name, dependency_hash).enforce
       end
 
+      def file_loader
```
WilliamMcCumstie: It gets used in #147, will look into the error

src/config.rb
```diff
@@ -63,11 +67,26 @@ def initialize(file = nil, options = {})
       end
     end
 
+    def loader
+      @loader ||= Validator::Loader.new(self)
+    end
+
+    def saver
+      @saver ||= Validator::Saver.new(self)
+    end
```
WilliamMcCumstie: Ok, I think at this point I only use the file loader built into the `BaseCommand` so it can be removed from the `Config`

src/validator/saver.rb
WilliamMcCumstie: It will get used in #147

src/validator/loader.rb
WilliamMcCumstie: Will make a trello card for it

src/file_path.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    private
+
+    attr_reader :config
+
+    def define_constant_paths
```
WilliamMcCumstie: It is used during `initialize` currently it gets indirectly tested. Can make it more explicit

src/validator/loader.rb
WilliamMcCumstie: This file maybe validated in the future however, will talk about it tomorrow

src/command_helpers/base_command.rb
WilliamMcCumstie: Error fixed

src/file_path.rb
WilliamMcCumstie: I don't think these methods should be defined in the `config`. At the moment we have paths that can be defined in `Config` and `Constants` which makes it hard to locate where a file is going to be. Also  `answer` files should always be loaded through the validator. So I think the better option is to remove the methods from `Config` and add them to `FilePath`.

src/validator/loader.rb
WilliamMcCumstie: Will make a card for it. Will be easier to make the change after #147 is merged

src/validator/loader.rb
bobwhitelock: If we validate it in the future we can always add it here then, no need to add extra indirection I think :slightly_smiling_face:

PR #147, opened 2017-08-02
Remove primary group command

WilliamMcCumstie:
This PR is based on the `ValidatorLoader`. Please merge #146 first.

Initially this command was going to be called `metal remove group`. However as we have discussed previously, the term `group` is ambiguous in metalware. Hence why the command has been renamed `metal remove primary_group` to better reflect its functionality. Should that be `metal remove primary-group`?

When removing a primary group, all nodes answers that belong to that primary group as well as the primary group answers are deleted. Also the primary group is removed from the group cache. Should we rename group cache to primary_group cache? Currently deleting a group will alter the `group_index` of other groups defined after it.

One caveat is how we handle additional groups. As other nodes could belong to the `group` but not the `primary_group` the expect behaviour becomes ambiguous. Due to the complexity of the task, I have decided to keep it simple, in that `remove primary_group` DOES NOT delete the `group`. This means it will still remain as an `additional_group` for other nodes, however it no-longer exists as a `primary_group`.

### DIFF
--- a/spec/commands/remove/group_spec.rb
+++ b/spec/commands/remove/group_spec.rb
@@ -22,7 +22,7 @@
 # https://github.com/alces-software/metalware
 #==============================================================================
 require 'filesystem'
-require 'commands/remove/primary_group'
+require 'commands/remove/group'
 require 'nodeattr_interface'
 require 'node'
 require 'config'
@@ -30,7 +30,7 @@ require 'ostruct'
 require 'validator/loader'
 require 'spec_utils'
 
-RSpec.describe Metalware::Commands::Remove::PrimaryGroup do
+RSpec.describe Metalware::Commands::Remove::Group do
   let :filesystem do
     FileSystem.setup do |fs|
       fs.with_minimal_repo
@@ -54,8 +54,7 @@ RSpec.describe Metalware::Commands::Remove::PrimaryGroup do
     filesystem.test do |_fs|
       answer_check = RSpecRemoveGroup::AnswerFileChecker
                      .new(self, config, primary_group, primary_nodes)
-      Metalware::Commands::Remove::PrimaryGroup
-        .new([primary_group], OpenStruct.new)
+      Metalware::Commands::Remove::Group.new([primary_group], OpenStruct.new)
       answer_check.check
       yield if test_block
     end
--- a/spec/commands/remove/group_spec.rb
+++ b/spec/commands/remove/group_spec.rb
@@ -57,7 +57,8 @@ RSpec.describe Metalware::Commands::Remove::Group do
   end
 
   def expected_deleted_files(group)
-    Metalware::NodeattrInterface.nodes_in_primary_group(group)
+    Metalware::NodeattrInterface
+      .nodes_in_primary_group(group)
       .map { |node| "nodes/#{node}.yaml" }
       .unshift(["groups/#{group}.yaml"])
       .map { |f| File.join(config.answer_files_path, f) }
--- a/spec/commands/remove/group_spec.rb
+++ b/spec/commands/remove/group_spec.rb
@@ -22,12 +22,95 @@
 # https://github.com/alces-software/metalware
 #==============================================================================
 require 'filesystem'
+require 'commands/remove/group'
+require 'nodeattr_interface'
+require 'config'
+require 'ostruct'
 
 RSpec.describe Metalware::Commands::Remove::Group do
   let :filesystem do
     FileSystem.setup do |fs|
       fs.with_minimal_repo
-      fs.with_answer_fixtures('answers/set1')
+      fs.with_answer_fixtures('setup1/answers')
     end
+  end
+
+  let :config { Metalware::Config.new }
+
+  def generate_node_list(prefix, max_index: 10)
+    [*1..max_index].map { |idx| "#{prefix}#{ "0" if idx < 10 }#{idx}" }
+  end
+
+  def run_remove_group(primary_group, primary_nodes)
+    allow(Metalware::NodeattrInterface).to \
+      receive(:nodes_in_primary_group).and_return(primary_nodes)
+
+    filesystem.test do |_fs|
+      answer_check = RSpecRemoveGroup::AnswerFileChecker
+                       .new(self, config, primary_group, primary_nodes)
+      Metalware::Commands::Remove::Group.new(primary_group, OpenStruct.new)
+      answer_check.check
+    end
+  end
+
+  context 'with no other groups' do
+    it 'removes group and node answer files' do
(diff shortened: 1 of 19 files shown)


bobwhitelock:
Rebased for easier reviewing

spec/nodeattr_interface_spec.rb
```diff
@@ -27,6 +27,23 @@
...
+    end
+
+    describe '#nodes_in_primary_group' do
+      it 'returns only the nodes in the primary group' do
+        nodes = Metalware::NodeattrInterface.nodes_in_primary_group('group1')
+        nodes_not_in_primary_group = nodes.select do |node|
+          node.match(/\AnodeA\d\d\Z/).nil?
+        end
+        expect(nodes.length).to eq(10)
+        expect(nodes_not_in_primary_group.length).to eq(0)
```
WilliamMcCumstie: I still like the regex check as checking the length is shorter doesn't guarantee the correct nodes have been selected. However instead of using a negative match and confirming it has a length of 0, I have switched it to use the built in rspec `match` method. This should make it a bit clearer.

bobwhitelock: I'm not sure I follow - I don't mention anything about checking the length is shorter. I'm just saying that if you want to check a method returns the correct array, it seems the simplest and clearest way to do so would be a single assertion like `expect(nodes).to eq(['node01', 'node02', 'node03'])`. I can't see a downside to doing this in this case, and any checking of the lengths of the array or that each element matches a regex would then be redundant and only make things less clear.

spec/commands/remove/group_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+
+  let :config { Metalware::Config.new }
+
+  before :each do
+    SpecUtils.mock_validate_genders_success(self)
+    SpecUtils.use_mock_genders(self, genders_file: 'setup1/genders')
+  end
+
+  def generate_node_list(prefix, max_index: 10)
+    [*1..max_index].map { |idx| "#{prefix}#{'0' if idx < 10}#{idx}" }
```
WilliamMcCumstie: Removed and replaced with `nodeattr`, forgot I had already tested it separately

spec/commands/remove/group_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+  end
+
+  def run_remove_group(primary_group, primary_nodes, &test_block)
```
WilliamMcCumstie: Renamed the method and moved the standard expects into it

spec/commands/remove/group_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+    filesystem.test do |_fs|
+      answer_check = RSpecRemoveGroup::AnswerFileChecker
+                     .new(self, config, primary_group, primary_nodes)
+      Metalware::Commands::Remove::Group.new([primary_group], OpenStruct.new)
+      answer_check.check
+      yield if test_block
```
WilliamMcCumstie: fixed

spec/commands/remove/group_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+        other_node_groups = Metalware::Node.new(config, 'nodeA01').groups
+        expect(other_node_groups).to include('group1')
+      end
+    end
+  end
+end
+
+# Can only be ran inside RSpec AND FileSystem
+module RSpecRemoveGroup
+  class AnswerFileChecker
```
WilliamMcCumstie: Removed and moved the methods inside of `rspec`

spec/nodeattr_interface_spec.rb
WilliamMcCumstie: It's just that their is 10 nodes in the group, so defining the array manually is a bit of a pain. Now we check that the array contains 10 elements and that those 10 elements match the correct regex. It takes all of 2 lines of code

bobwhitelock: Yep - but the other part of my original comment was that I can't see why we need so many nodes in the group. The tests would be as good, if not better as they would be clearer, with just 2 or 3 nodes in each group, unless there's some reason I've missed why we need more.

WilliamMcCumstie: Ahh, that's because it is the same setup as I used for removing a group which was generate using the answer files. So it isn't as easy as editing a single file.

WilliamMcCumstie: So just tried using `testnodes` in the default genders file. However in order to test it I needed to add new nodes in a different primary group to the `testnodes` gender, which caused a bunch of other tests to break.

So the only way I see to test this the way you are suggesting is to add two new gender groups specifically for this test. Up to you?

spec/commands/remove/group_spec.rb
```diff
@@ -40,29 +40,42 @@
...
-  def run_remove_group(primary_group, primary_nodes, &test_block)
+  def test_remove_group(group)
     filesystem.test do |_fs|
-      answer_check = RSpecRemoveGroup::AnswerFileChecker
-                     .new(self, config, primary_group, primary_nodes)
-      Metalware::Commands::Remove::Group.new([primary_group], OpenStruct.new)
-      answer_check.check
-      yield if test_block
+      Metalware::Commands::Remove::Group.new([group], OpenStruct.new)
+      expect(expected_deleted_files(group)).to include(*deleted_files)
```
bobwhitelock: Thanks for making these changes, this looks clearer to me now. Can this line use `eq` rather than `include`? As these should be exactly the same shouldn't they?

spec/nodeattr_interface_spec.rb
bobwhitelock: You're [already using a genders file just for this test](https://github.com/alces-software/metalware/pull/147/files#diff-f3cb3abf9ef9f2c5c95bd109c95804a0R32) - can you change this to specify nodes as `01-03` or `01-02`, and then specify the node names directly in the test? You'll need to change some other tests where the same genders file is used to use the same range, but this shouldn't be a complicated change afaics.

spec/commands/remove/group_spec.rb
WilliamMcCumstie: No they are not exactly the same. `expected_deleted_files` is every file that should be deleted including files that don't exist in the first place. Where as `deleted_files` are the files only include the files that existed in the first place.

bobwhitelock: OK, thanks for explaining

spec/nodeattr_interface_spec.rb
WilliamMcCumstie: Ok I have adjusted `setup1` and implemented the change.

PR #148, opened 2017-08-02
Improve build interface handling and getting deployment server IP

bobwhitelock:
This PR resolves https://trello.com/c/zJLDMucI; it is based on https://github.com/alces-software/metalware/pull/144.

This PR adds handling for a new config file, `server.yaml`, for properties of the Metalware setup and deployment server itself. This file is rendered from the repo with each `metal configure` command, prior to the other domain templates (`hosts` and `genders`); before being saved it is validated to check a valid `build_interface` has been specified, as this must be an interface which exists on the deployment server or future template rendering will break. The rendered file is then loaded when forming the templating namespace for rendering other files, with the specified `build_interface` used to determine the IP of the Metalware server on the build network.

For a full explanation of why I took this approach see https://github.com/alces-software/metalware-default/pull/9.


WilliamMcCumstie:
Rebased on develop

src/domain_templates_renderer.rb
```diff
@@ -12,27 +13,53 @@ def initialize(config, genders_invalid_message: nil)
...
       render_template(
         server_config_template,
         to: Constants::SERVER_CONFIG_PATH
-      )
+      ) do |rendered_config|
+        validate_rendered_server_config(rendered_config)
+      end
+    end
+
+    def validate_rendered_server_config(rendered_config)
```
WilliamMcCumstie: This is fine for now, but we may want to extract this into the `Validator` module if it needs to be accessed from elsewhere

bobwhitelock: Yep, it may be worth adding a fuller validation of the rendered `server.yaml` at some point. At the moment we just need to check a  value

PR #149, opened 2017-08-02
Default hunter interface to build interface

bobwhitelock:
This PR is a small addition, based on https://github.com/alces-software/metalware/pull/148. Now that we define the Metalware build interface in a consistent location, the rendered `server.yaml` file, this PR makes this build interface be the default interface used when `metal hunter` is run, rather than defaulting to `eth0` as we did previously. This is useful as this is almost always (or possibly just always) the interface we want to hunt on, so specifying this as the default means we should normally not need to pass this option.


WilliamMcCumstie:
Rebased on develop

WilliamMcCumstie:
LGTM

PR #150, opened 2017-08-02
Install gethostip in installer

bobwhitelock:
This is needed by Metalware but was not being installed when installing Metalware on a machine running a minimal Centos install. Based on https://github.com/alces-software/metalware/pull/149; fixes https://trello.com/c/wga3bOlt.


WilliamMcCumstie:
rebased on develop

PR #151, opened 2017-08-03
Privatise all the things

bobwhitelock:
We only ever instantiate our command classes, and don't run any other methods from outside the classes, so we can just make everything in these private. Based on https://github.com/alces-software/metalware/pull/150.


WilliamMcCumstie:
Rebased on develop. LGTM

PR #153, opened 2017-08-07
Allow build template selection in configs (closes #143)

### LINKED ISSUE #143: PXE and Kickstart File Selection in Configs
Currently, to specify non-default PXE and kickstart files for a node/group, arguments to `metal build` are required. While this works it could be more flexible to allow for groups to contain nodes which are built from different PXE/kickstart files. The idea here is to move the definitions of these files to the configs, such that `domain.yaml` would contain lines like:

```
build:
  pxe: default
  kickstart: default
```

bobwhitelock:
This PR allows the PXELINUX and Kickstart templates used when building each node to be specified in the repo config for that node; following the meeting earlier the `--kickstart` and `--pxelinux` options have also been removed as they are now superfluous.

Trello: https://trello.com/c/ySxv4Z1s/155-pxe-and-kickstart-file-selection-in-configs. Corresponding repo change demonstrating use of this feature: https://github.com/alces-software/metalware-default/commit/00c0bf0f47cd47a6d42fbb46a91ccb410fbbc976.

### DIFF
--- a/src/commands/build.rb
+++ b/src/commands/build.rb
@@ -175,11 +175,11 @@ module Metalware
           nodes.select do |node|
             !rerendered_nodes.include?(node) && node.built?
           end
-                .tap do |nodes|
+               .tap do |nodes|
             render_permanent_pxelinux_configs(nodes)
             rerendered_nodes.push(*nodes)
           end
-                .each do |node|
+               .each do |node|
             Output.stderr "Node #{node.name} built."
           end
 
--- a/src/commands/build.rb
+++ b/src/commands/build.rb
@@ -109,7 +109,7 @@ module Metalware
       end
 
       def render_kickstart(parameters, node)
-        kickstart_template_path = template_path :kickstart
+        kickstart_template_path = template_path :kickstart, node: node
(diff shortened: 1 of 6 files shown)


WilliamMcCumstie:
rebased on develop

src/commands/build.rb
```diff
@@ -55,15 +55,28 @@ def run
...
+            full_template_path = template_path(template_type, node: node)
+            repo_relative_path_to(full_template_path)
+          end
+        end.flatten.uniq
+      end
+
+      def repo_relative_path_to(path)
+        repo_path = Pathname.new(config.repo_path)
+        Pathname.new(path).relative_path_from(repo_path).to_s
+      end
```
WilliamMcCumstie: This method should be moved into `FilePath` class so it is available to other methods

bobwhitelock:
Thanks for the review and suggestion, I've made that change. Can you merge this if you're happy with it? :slightly_smiling_face:

WilliamMcCumstie:
Thanks for that change LGTM

PR #155, opened 2017-08-09
Various build command and tests refactoring

bobwhitelock:
This PR contains various refactoring done prior to implementing the PR I'm about to base off this. There shouldn't be any new functionality here, just various tweaks and refactorings made prior to the next branch; see individual commits for details. This is based off https://github.com/alces-software/metalware/pull/153.

### DIFF
--- a/spec/minimal_repo.rb
+++ b/spec/minimal_repo.rb
@@ -25,12 +25,9 @@
 
 module MinimalRepo
   class << self
-    DIRECTORIES = [
-      '.git',
-      'files',
-    ].freeze
-
     FILES = {
+      '.git/': nil,
+      'files/': nil,
       'pxelinux/default': "<%= alces.firstboot ? 'FIRSTBOOT' : 'NOT_FIRSTBOOT' %>\n",
       'kickstart/default': '',
       'hosts/default': '',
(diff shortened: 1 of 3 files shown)


spec/minimal_repo.rb
```diff
@@ -5,13 +5,7 @@ module MinimalRepo
   class << self
     DIRECTORIES = [
       '.git',
-      'config',
-      'hosts',
-      'genders',
-      'dhcp',
       'files',
```
WilliamMcCumstie: Do we still need to have this array if it only has `.git` and `files`? Can this be done the same way as `config`, `hosts`, etc?

bobwhitelock: We need it for those as we only want to create an empty directory in those cases. I guess we could specify a file we don't actually want in those directories and then have `create_files_at` implicitly create the directory as well, but it seems that would make things slightly confusing in order to remove a couple of lines of code.

bobwhitelock: Actually I may be able to do this simply, let me just try it out and see if it looks nicer.

bobwhitelock: I've made this change, let me know what you think: https://github.com/alces-software/metalware/commit/86b15e8acc04ca284e520d5db5f6b7091b135dec.

bobwhitelock:
Made the change, and I've also rebased this on the latest `develop` to resolve the conflicts with the recently merged branch.

WilliamMcCumstie:
Thanks for that change, much simpler now

PR #157, opened 2017-08-10
Create static group indexing

WilliamMcCumstie:
Now that a group can be removed, it is important to insure that the group indexing does not change as a result. This is done by saving the next index to use in the group cache and assigning each group a static index.

In order to implement the change, a new `GroupCache` class has been created to control access to the cached file. First the `GroupCache` was created to mimic the current usage. Then all methods that needed to access the cache where switched over the new object.

Finally the underlining data structure of the cache was changed and a few broken tests needed to be updated.


bobwhitelock:
Rebased as first 2 commits were already in `develop` from merging a previous branch.

spec/group_cache_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+  let :config { Metalware::Config.new }
+  let :cache { Metalware::GroupCache.new(config) }
+
+  let :filesystem do
+    FileSystem.setup do |fs|
+      fs.with_group_cache_fixture('cache/groups.yaml')
+    end
+  end
+
+  it 'checks if a group has been configured' do
```
bobwhitelock: For each of these tests, if you wrap them in a block like `describe '#group?' do` then it will make it clearer which tests are for which method on the class.

src/commands/configure/group.rb
```diff
@@ -24,29 +47,12 @@ def answers_file
...
-          primary_groups.include? group_name
-        end
-
-        def primary_groups
-          groups_cache[:primary_groups] ||= []
-        end
-
-        def groups_cache
-          @groups_cache ||= Data.load(Constants::GROUPS_CACHE_PATH)
+          cache.add(group_name) unless cache.group?(group_name)
```
bobwhitelock: Rather than asking the cache if it already contains the group, and then adding it if it doesn't, can we just tell it to add the group and let it decide if it should record it? I.e.: can we just do `cache.add` here and have that contain the conditional? I think this is a situation where it would be useful to follow [tell don't ask](https://robots.thoughtbot.com/tell-dont-ask).

WilliamMcCumstie:
Thanks for the review, I have made those changes and add a few more index checks in the tests

bobwhitelock:
Thanks for the changes, I've just made a couple of minor additions to the tests.

PR #156, opened 2017-08-09
Add basic metal build method (closes #145)

bobwhitelock:
This PR adds a `basic` `metal build` method for nodes, specified by the `build_method key in the repo config for a node, which is an alternative to the existing default Kickstart build method. See https://github.com/alces-software/metalware/commit/3fb444b94ca656e3216847fdb472cf15c7fa02c7 for the main details of what is involved in this method.

This resolves https://github.com/alces-software/metalware/issues/145 / https://trello.com/c/0wZYT7o9/164-add-basic-metal-build-method. It is based on https://github.com/alces-software/metalware/pull/155.

PR #158, opened 2017-08-10
Only rerender parts of hosts and genders

bobwhitelock:
This PR changes how we handle rendering the `hosts` and `genders` files, to only render part of each file (which I call the 'managed section') and keep the rest of the file outside this section the same. This change will allow cluster admins to edit these files manually outside of this section, without needing to know or care about Metalware or how it changes these files.

An exact description of how this new rendering method functions depending on the existing file contents is in 3887a39d93ce2359b23dcc569aeffbe6e31f1ce0.

Based on #156; Trello: https://trello.com/c/pb8yT6qP.

### DIFF
--- a/src/templater.rb
+++ b/src/templater.rb
@@ -28,6 +28,7 @@ require 'active_support/core_ext/string/strip'
 require 'constants'
 require 'metal_log'
 require 'exceptions'
+require 'utils'
 require 'templating/iterable_recursive_open_struct'
 require 'templating/missing_parameter_wrapper'
 require 'templating/magic_namespace'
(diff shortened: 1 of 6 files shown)


WilliamMcCumstie:
Rebased on develop

bobwhitelock:
@WilliamMcCumstie: I've not finished making the changes to the previous PR yet, you may want to wait for those first.

WilliamMcCumstie:
I have one of the build tests failing on my machine and Travis. Looks like something is timing out. Will do some digging why

bobwhitelock:
I've rebased this on the latest https://github.com/alces-software/metalware/pull/156 to fix the issue with the previous rebase.

WilliamMcCumstie:
Rebased on develop

src/templater.rb
```diff
@@ -42,6 +42,17 @@ class Templater
     # `metal configure` commands.
     EOF
 
+    MANAGED_START = 'METALWARE_START'
+    MANAGED_END = 'METALWARE_END'
+    MANAGED_START_COMMENT = "# #{MANAGED_START}"
+    MANAGED_END_COMMENT = "# #{MANAGED_END}"
```
WilliamMcCumstie: Could we make these comments stand out a bit more? If the managed section is long, then it could be easily missed when scrolling through.
```
Currently:
# METALWARE_START

Maybe:
######### METALWARE_START #########
```

bobwhitelock:
@WilliamMcCumstie: that's a good point; I've made that change. Thanks for the review, I'm assuming this is good to merge now.

PR #159, opened 2017-08-10
Renamed Validator to Validation

WilliamMcCumstie:
Renamed the `Validator` files and module to `Validation`.

Their are a few local variables called `validator`, but it is probably not worth changing them.

Also renamed the file loader `configure` method to `configure_data`

Note: Their is a build test failing, but it also fails in #158. Likely the failure has resulted from one of the merges with `develop`


bobwhitelock:
Rebased on `develop`; I've squashed the middle two commits as we need to change the `require`s at the same time as we move the files otherwise many tests break.

PR #160, opened 2017-08-10
Remove IO.abort and replace it with raise

WilliamMcCumstie:
IO.abort has been removed as it causes the error message to skip the standard Metalware logging, making bugs harder to track. It has been replaced with raising Metalware::DomainTemplatesInternalError

The error messaged used for this error is reasonable generic as it is determined from the name of the error method. This can be improved in the future.


bobwhitelock:
Thanks!

PR #162, opened 2017-08-14
Make group_index always default to 0 (fixes #161)

### LINKED ISSUE #161: `alces.group_index` is not an integer
`alces.group_index` does not function the same way as `alces.index`. Take for example the following line from the domain config:
```
10.10.<%= alces.group_index %>.<%= alces.index + 1 %>
```
This will work absolutely fine, however if it is modified to:
```
10.10.<%= alces.group_index + 1 %>.<%= alces.index + 1 %>
```
Then the configure bombs out with the following error:
```
error: Attempted to call method `+` of unset template parameter. Use --trace to view backtrace
```
This issue can be fixed by setting the line to:
```
10.10.<%= alces.group_index.to_i + 1 %>.<%= alces.index + 1 %>
```

So, it seems that `alces.group_index` needs to be an integer so that the behaviour matches that of `alces.index`

bobwhitelock:
Rather than having situations where this can be nil, which appears to just make templating more difficult without providing any benefit. Fixes issue discussed in https://github.com/alces-software/metalware/issues/161.

### DIFF
--- a/spec/fixtures/cache/groups.yaml
+++ b/spec/fixtures/cache/groups.yaml
@@ -1,4 +1,4 @@
-next_index: 2
+next_index: 3
 primary_groups:
-  some_group: 0
-  testnodes: 1
+  some_group: 1
+  testnodes: 2
--- a/spec/fixtures/setup1/cache/groups.yaml
+++ b/spec/fixtures/setup1/cache/groups.yaml
@@ -1,6 +1,6 @@
 ---
-next_index: 3
+next_index: 4
 primary_groups:
-  group1: 0
-  group2: 1
-  nodes: 2
+  group1: 1
+  group2: 2
+  nodes: 3
--- a/spec/group_cache_spec.rb
+++ b/spec/group_cache_spec.rb
@@ -52,7 +52,7 @@ RSpec.describe Metalware::GroupCache do
         cache.add('new_group')
         expect(cache.group?('new_group')).to eq(true)
 
-        next_available_index = 2
+        next_available_index = 3
         expect(cache.index('new_group')).to eq(next_available_index)
       end
     end
@@ -95,13 +95,13 @@ RSpec.describe Metalware::GroupCache do
       filesystem.test do
         group1 = 'group1'
         cache.add(group1)
-        expect(cache.index(group1)).to eq 2
+        expect(cache.index(group1)).to eq 3
 
         cache.remove(group1)
 
         group2 = 'group2'
         cache.add(group2)
-        expect(cache.index(group2)).to eq 3
+        expect(cache.index(group2)).to eq 4
       end
     end
   end
--- a/spec/node_spec.rb
+++ b/spec/node_spec.rb
@@ -151,16 +151,10 @@ RSpec.describe Metalware::Node do
   describe '#group_index' do
     let :filesystem { FileSystem.setup }
 
-    def expect_group_index_raises_for_node(node)
-      filesystem.test do
-        expect { node.group_index }.to raise_error(
-          Metalware::UnconfiguredGroupError
-        )
-      end
-    end
-
-    it 'raises when groups.yaml does not exist' do
-      expect_group_index_raises_for_node(testnode01)
+    it 'returns 0 when groups.yaml does not exist' do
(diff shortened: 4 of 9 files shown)


bobwhitelock:
We don't really want to access `group_index` if the group doesn't exist - but when rendering a template like `hosts`, which is rendered for the entire domain without a particular node context, the config is first loaded for the domain. If `alces.group_index` is `nil` here then in a config where it is added to, like @ColonelPanicks was doing in the original issue, would error, even though we don't actually care about the final config value.

I believe you're mistaken in your suggestion in https://github.com/alces-software/metalware/issues/161#issuecomment-322239653, I think your recent change to the groups cache is fine; the issue here is how @ColonelPanicks is now using `alces.group_index` (which is a completely reasonable way to want to use it, we just weren't adding anything to it before so didn't notice this issue before).

We can use an index of `0`, although this does overload the meaning of an `alces.group_index` of `0`, this is consistent with how `alces.index` works so is arguably a less complicated/confusing way for things to work. 

Regarding the suggested alternate fixes:

1. This is what we did already, and is what was causing this issue, although we did catch the error and return `nil` in the `MagicNamespace` so templating didn't blow up in this case.

2. (and 3 also) Either of these would also work, although both would be less consistent with how `alces.index` works. Since we shouldn't actually be using the node context `group_index` when not rendering a template in a node context, I don't think which we use really matters much.

I think the truly robust fix for this sort of issue would be to overhaul how we handle templating to have rendering domain templates not just be a special case of rendering for a node with the node name set to `nil`, and so we would not try to form a meaningless node IP using a meaningless `group_index` when loading the config for the domain. There's various ways we could do this but this would be a big change without a massive benefit right now. Instead, I thought the simplest/best option right now was to make `group_index` be consistent with `index`, and give it a valid default value when we're not rendering for a specific node (which shouldn't actually be used in practise).

Let me know if you have any more thoughts on this, or we can discuss tomorrow if you like :slightly_smiling_face:.

WilliamMcCumstie:
Ahh right, I forgot that when the `hosts` template is constantly re-rendered. I think in that case this fix should suffice. It would be nice if we used a more robust method for `group_index` however I do agree that it is a lot of work with very little gain. Could you add a card for this?

For the time being, we need to resolve the conflict over `group_index` 0. Now, their is no point to try and make it act exactly like `index` because as soon as a node is removed, that `group_index` is dead. Hence I have no issue skipping an index. Also as we are trying to make something that doesn't break `hosts` we shouldn't use a negative index. That what `hosts` may not break.

So to fix it, could you change the starting `group_index` to 1 (should be in `GroupCache`). Also as this is not desirable behaviour, can you add a warning when either a `group_index` of 0 is used? This is important because that index is liable to change at anytime and thus might break something.

ColonelPanicks:
I think it's worth keeping `group_index` consistent with `alces.index`. Arguably, `alces.index` defaulting to 0 could cause conflicts as well, however, the default repo config sets `<%= alces.index + 1 %>` for IP addresses and the idea is that we'd set the group index to be similar (`<%= alces.group_index + 1 %>`). If we start both indexes at differing values I can only see that becoming confusing so if we're going to start the group index at 1, can we also do that for the node index?

bobwhitelock:
@WilliamMcCumstie: I've bumped the `group_index` to start from 1; I can see how there is value in having a distinct index for the first group vs when we have no group. We can't have a warning if this is used however as, as pointed out in the original issue, this will always be used when rendering a template for no particular node even though we don't use the final config values; so we would always get this warning when rendering any domain templates, making it effectively useless.

@ColonelPanicks: That sounds reasonable to me too, for the same reason and to be consistent; I've also made that change.

Everybody happy for this to be merged now?

ColonelPanicks:
I'll be sure to make relevant changes to `metalware-default` to reflect these index changes.

bobwhitelock:
@ColonelPanicks: Thanks, you should just be able to look at https://github.com/alces-software/metalware-default/pull/10 and let me know if I've missed anything :slightly_smiling_face:.

WilliamMcCumstie:
LGTM @bobwhitelock. Just added one additional commit to fix the cache file stored in the fixtures. Otherwise looks good to me

PR #163, opened 2017-08-15
Tidy up command args and options handling

bobwhitelock:
This PR contains some tidying up of how we handle `args` and `options `in Metalware in some places. The important change here is https://github.com/alces-software/metalware/commit/9b4eb1ca1b82a993b29fbd297fb11450907acda9, which I made so I could access the `options` in multiple `configure` commands without needing to independently set these in their different `setup` methods (the details of this are unimportant; it will be in a future PR). The other changes came out of this, when I spotted a few places we could improve things; see commits for details.


WilliamMcCumstie:
LGTM

PR #165, opened 2017-08-15
Implement `named` DNS resolution (close #108)

### LINKED ISSUE #108: Named template management tool
I'd had this down as a future feature but logging it now because it might be more relevant now with regards to doing more persistent hosts files functionality as its quite a similar problem.

Currently metalware uses a host file distributed around the deployed nodes at build time in order to provide basic DNS resolution around a platform. Thats OK for basic clusters but in general, we switched to doing proper DNS on our managed clusters a few years ago because its considerably more flexible and resilient, its generally slower but we havn't really seen any problems in the last few years..

So as I see it, the basic use case would be...

Some templates in metalware-default for each network (forward and reverse)..
eg..
forwards...
```
$TTL 300
@                       IN      SOA     <%=hostip%>. nobody.example.com. (
                                        <%= getrubytomakeatimestring %>   ; Serial
                                        600         ; Refresh
                                        1800         ; Retry
                                        604800       ; Expire
                                        300          ; TTL
                                        )

                        IN      NS      $cobbler_server.
@       IN MX   <%=hostip.quad4 %>  <%=hostip%>.<%=domain%>

IN NS <%=hostip%>.

<% for each machine with a private network IP %>
<%=hostname%> IN A <%=ip%>;
<% end %> 
```
reverse...
```
$TTL 300
@                       IN      SOA     <%=hostip%>. nobody.example.com. (
                                        <%= getrubytomakeatimestring %>   ; Serial
                                        600         ; Refresh
                                        1800         ; Retry
                                        604800       ; Expire
                                        300          ; TTL
                                        )

                        IN      NS      <%=hostip%>.
<% for each machine with a private network IP %>
<%=ip.quad4%>.<%ip.quad3%>    IN PTR  <%=hostsFQDNonthisnetwork%>.;
<% end %>
```

Each time a new host / group of hosts are added, these templates (for all networks) would need to be regenerated and named restarted to pick up the changes..

Theres more I could add (the setup of named etc) but thats pretty much what we'd actually need to get metalware to do dynamically. Hopefully gives you guys an initial idea of the use case so we can discuss. 

Mazin - let me know your thoughts..

WilliamMcCumstie:
Fixes #108

Instead of only using `hosts` for DNS, it is now possible to use `named` instead. This is done by specifying `dns_type` in the `domain.yaml` config file. Currently only two DNS types are valid: `hosts` and `named`. If the `dns` type is not given, it reverts to `hosts` and issues a warning.

If `named` isn't already running, it runs a setup method. Here it renders a `/etc/named.conf` template which is stored in metalware itself. This template includes `/etc/named/metalware.conf` and thus does not need to be re-rendered each time.

Then when ever the `hosts` template would have otherwise been rerendered, the following templates are rendered instead: `metalware.conf`, the forward templates and the revers templates.Rendering `metalware.conf` is straight forward enough, making it a reasonable standard implementation. 

However the forward and reverse templates are trickier. Firstly the save file name is not predetermined and have to be supplied from the config file. This is done by including a `named_fwd_zone` and `named_rev_zone` in each of the `networks` contained in repo configs `domain.yaml`. If these are not included, the file is not rendered, however an warning is NOT issued as not all networks need it.

The next issue is that multiple forward and reverse files are needed for each network. This was done previously by rendering a script that did this. However the looping functionality has been moved into `Metalware` so the user doesn't need to worry about it. Instead they only need to give the content of what is going to be in each file.

The catch is they will not know the `zone` and `net` information for the network. Previously this was given by the fact it was defined in the loop, but this is no longer the case. Instead a `alces_named` variable is passed to erb which contains `.net` and `.zone` to access these parameters.

It might be a good idea to move this into the magic namspace so they can be accessed with `alces.named.zone`. 

Sorry for the long PR, had a few design decisions I had to make.


bobwhitelock:
@WilliamMcCumstie: the same test's failing for me locally and on my controller; looks like a file isn't being loaded into the `FakeFS`. Does it not fail for you?

WilliamMcCumstie:
Ahh thought It was an install issue. But while I was fixing that, I forgot to update the tests. Should be working now

src/dns/named.rb
```diff
@@ -0,0 +1,141 @@
...
+
+      def render_zone_template(direction, zone, net)
+        zone_name = case direction
+                    when 'forward'
+                      net[:named_fwd_zone]
+                    when 'reverse'
+                      net[:named_rev_zone]
+                    end
+        zone_hash = named_zone_hash(zone, net)
+        return unless zone_name
```
bobwhitelock: Don't think we need this line? The tests pass without it, and afaics it shouldn't be reached. If we do reach it by accident in future though, it would be better to fail noisily because this is `nil` rather than silently I think.

src/dns/named.rb
```diff
@@ -0,0 +1,141 @@
...
+        SystemCommand.run(RESTART_NAMED_CMDS)
+      end
+
+      def each_network
+        repo_config[:networks]&.each { |data| yield data }
+      end
+
+      def named_zone_hash(zone, net)
+        {
+          alces_named: {
```
bobwhitelock: Can this be merged in to the magic namespace when we're templating? I think this would be nicer than having it magically appear in the main templating namespace when rendering these templates. I think you should be able to handle this similarly to `firstboot` for build templates. Possibly it's be easier to do this in a future PR though, since your next PR makes a lot of changes to how we do templating.

templates/named.conf.erb
```diff
@@ -0,0 +1,29 @@
...
+          recursion yes;
+
+
+          dnssec-enable no;
+          dnssec-validation no;
+          dnssec-lookaside auto;
+
+          forward first;
+          forwarders {
+              <%= external_dns %>;
```
ColonelPanicks: Had noted down this mismatch on the [original issue](https://github.com/alces-software/metalware/issues/108#issuecomment-323026684)

bobwhitelock: Thanks for the link @ColonelPanicks. @WilliamMcCumstie do we need to do anything about the second point there:

> Create an empty file at /etc/named/metalware.conf so that the file exists and named doesn't fail out

I don't understand why `named` would fail though, as shouldn't this file always be rendered?

WilliamMcCumstie: Yeah I will need to implement that. Their is a include in the bash script, so it will fail if the file doesn't exist

WilliamMcCumstie: It doesn't get rendered until the first run, so on setup in needs to be created otherwise setup fails

src/domain_templates_renderer.rb
```diff
@@ -123,6 +128,26 @@ def display_genders_error(nodeattr_error)
...
+      case repo_config.dns_type
+      when nil
+        MetalLog.warn(MISSING_DNS_TYPE)
+        render_hosts
+      when 'hosts'
+        render_hosts
+      when 'named'
+        update_named
+      else
+        raise InvalidInput, "Invalid DNS type: #{repo_config.dns_type}"
```
WilliamMcCumstie: Fixed

src/dns/named.rb
```diff
@@ -0,0 +1,141 @@
...
+      end
+
+      # TODO: These commands will break hosts DNS. Might be a good idea to run
+      # similar commands for hosts
+      RESTART_NAMED_CMDS = <<~EOF.strip_heredoc
+        systemctl disable dnsmasq
+        systemctl stop dnsmasq
+        systemctl enable named
+        systemctl restart named
+      EOF
```
WilliamMcCumstie: updated

src/dns/named.rb
```diff
@@ -0,0 +1,141 @@
...
+
+      # DO NOT RENAME, DANGEROUS FOR TESTS
```
WilliamMcCumstie: Removed

spec/dns/named_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  let :correct_base_named_conf do
+    externaldns = named.send(:repo_config)[:externaldns]
+    "DNS: #{externaldns}"
+  end
+
+  context 'without a setup named server' do
+    before :each do
+      allow(named).to receive(:setup?).and_return(false)
+      # Prevents restart_named from running, DANGEROUS
```
WilliamMcCumstie: updated

src/dns/named.rb
WilliamMcCumstie: Removed

src/file_path.rb
```diff
@@ -81,8 +86,13 @@ def template_file_name(template_type, node:)
     end
 
     def repo_template(template_type, node:)
-      repo_specified_templates = node.repo_config[:templates] || {}
-      repo_specified_templates[template_type]
+      templates = if node.nil?
```
WilliamMcCumstie: fixed

templates/named.conf.erb
WilliamMcCumstie: Fixed both the `externaldns` issue and the creation of the file as specified in #108

WilliamMcCumstie:
@bobwhitelock, made those fixes. I have added away to inject additional parameters into the alces namespace for individual templates. Should work for now.

src/dns/named.rb
WilliamMcCumstie: 28dbf63

src/domain_templates_renderer.rb
WilliamMcCumstie: 9c5b83e

src/dns/named.rb
WilliamMcCumstie: 9c5b83e

spec/dns/named_spec.rb
WilliamMcCumstie: 9c5b83e

templates/named.conf.erb
WilliamMcCumstie: dc139df and 8742c85

src/dns/named.rb
WilliamMcCumstie: 0553cb9

src/dns/named.rb
WilliamMcCumstie: 0a666a9

src/file_path.rb
WilliamMcCumstie: 4acee26

WilliamMcCumstie:
I haven't tested the changes to `SystemCommand`, stu was going to test them in the morning. But looking over the `Open3` documentation again, I think you might be right. Will change the syntax

WilliamMcCumstie:
As far as the `additional_parameters` go, we shouldn't add an additional input to `MagicNamespace` every time we have a one off additional parameter. `MagicNamespace` already has 5 inputs, so I didn't want to add a 6th. 

So instead I have generalised a way to add additional parameters into the namespace as their is a good chance we may want to do this again.

WilliamMcCumstie:
Rebased on develop

spec/dns/named_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  let :correct_base_named_conf do
+    externaldns = named.send(:repo_config)[:externaldns]
+    "DNS: #{externaldns}"
+  end
+
+  context 'without a setup named server' do
+    before :each do
+      allow(named).to receive(:setup?).and_return(false)
+      # Prevents restart_named from running as the bash commands can't be mocked
```
WilliamMcCumstie: fixed aa6e06e32ef1b828e5ef53e447e689bef7421d22

spec/dns/named_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+        rendered_named_conf = File.read(file_path.base_named).chomp
+        expect(rendered_named_conf).to eq(correct_base_named_conf)
+      end
+    end
+  end
+
+  context 'with a setup named server' do
+    before :each do
+      allow(named).to receive(:setup?).and_return(true)
+      # Prevents restart_named from accidentally running, DANGEROUS
```
WilliamMcCumstie: fixed aa6e06e

spec/dns/named_spec.rb
WilliamMcCumstie: Probably good for now, I will get stu to check the bash commands work in full anyway.

bobwhitelock:
Thanks for the new changes, this looks good now. Re:

> As far as the additional_parameters go, we shouldn't add an additional input to MagicNamespace every time we have a one off additional parameter. MagicNamespace already has 5 inputs, so I didn't want to add a 6th.
> 
> So instead I have generalised a way to add additional parameters into the namespace as their is a good chance we may want to do this again.
> 

this is fair enough, and it's generally reasonable to not keep adding new parameters to a class in this kind of situation. Having the parameters available within the `MagicNamespace` as explicit methods does make it clear what can be accessed within the class, which possibly useful, but generalising how these are passed in also seems reasonable. Either way we should do the same for all these sort of parameters, so I've added https://trello.com/c/hgyhK5RH to track this.

We should also document the `alces.named` stuff in the templating documentation; I've added this to https://trello.com/c/CvS8Yl5A to track this.

PR #166, opened 2017-08-18
Render repo config values on a as needed basis

WilliamMcCumstie:
Sorry for the long Pr, however their was quite a lot of work required to switch out how we use the config values in the `Templater`. Most of the commits are fairly small and deal with making the `Binding` work how it did previously. Alternatively they are fixing broken tests.

The easiest way to review this PR is step through the commits and read the messages. They describe the purpose of each step reasonable in depth.

In short, the `MissingParameterWrapper` has been replaced with a `Binding::Wrapper` and `RepoConfigParser` has been replaced with `Binding::Parameter`. The `MagicNamespace` is now created inside of the `Parameter` object. The big difference is config values are rendered when they are pulled out of the `IterativeRecursiveOpenStruct`.

Also, If you want to access a single value in the config, you can build a `Wrapper` object and query it directly. This then will return the value associated with that value without neededing to render the entire config.

In the process of switching refactoring I have ran into an issue with the answers forming an infinite recursion between the `Config` and the `Answers`. The fix for this is to render the answers on an as needed basis as well. However this should be implemented after this PR has been merged.


WilliamMcCumstie:
This PR as identified issues on how we define the context that the magic namespace operates at. At this point, getting the config parameters to be rendered on an as needed basis is unworkable. To implement the change would require an entire rework of the `MagicNamespace`.

The MagicNamespace currently works to magically which is hard to replicate. What probably would be require is the construction of a `DSL` instead.

Hence this PR will be ditched, and any small relevant changes will be made in a seperate PR.

bobwhitelock:
@WilliamMcCumstie: Oh, I'm part way through reviewing this, but not done yet as it's pretty dense. I'm not entirely sure what you mean, but sounds like you've run into some issues. Seems a shame to throw away this entirely but I'll hold off on the reviewing for now; why don't we get together tomorrow and we can discuss this more? I think it will be easier to discuss in person.

PR #169, opened 2017-08-21
Split build command with edit flags

WilliamMcCumstie:
First commit to implement the `edit` command. This commit concerns itself with how the `edit` command is used with `build`. At this point in time, the easiest way to implement it, is to split the build command into a render then build stages. This can be done with the use of flags on command line. At a latter date we can implement a more durable way to render the templates.

### DIFF
--- a/src/cli_helper/config.yaml
+++ b/src/cli_helper/config.yaml
@@ -155,7 +155,7 @@ commands:
       - *group_option
       - tags: [--edit-start]
         description: >
-          Only renders the build templates to be editted. Does not wait for the
+          Only renders the build templates to be edited. Does not wait for the
           nodes to be built
       - tags: [--edit-continue]
         description: >
--- a/src/cli_helper/config.yaml
+++ b/src/cli_helper/config.yaml
@@ -153,6 +153,14 @@ commands:
     action: Commands::Build
     options:
       - *group_option
+      - tags: [--edit-start]
+        description: >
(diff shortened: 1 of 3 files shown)


src/commands/build.rb
```diff
@@ -32,24 +32,27 @@
...
+        setup_edit_mode
         node_identifier = args.first
         @group_name = node_identifier if options.group
         @nodes = Nodes.create(config, node_identifier, options.group)
       end
 
       def run
-        render_build_templates
+        render_build_templates unless edit_continue
+        return if run_edit_start
```
bobwhitelock: Rather than having a conditional inside this conditional, it could be clearer to do something like this:

```ruby
if edit_start
  puts EDIT_START_MESSAGE
  return
end
```

You could make this a one-liner which does the same thing if you prefer, but either seems a bit clearer than checking `edit_start` twice in two places to me.

src/cli_helper/config.yaml
```diff
@@ -153,6 +153,14 @@ commands:
     action: Commands::Build
     options:
       - *group_option
+      - tags: [--edit-start]
+        description: >
+          Only renders the build templates to be editted. Does not wait for the
```
bobwhitelock: Should be `edited` here and below.

WilliamMcCumstie:
Fixed those issues

bobwhitelock:
Thanks!

PR #173, opened 2017-08-22
Implement the edit command

WilliamMcCumstie:
Fairly small PR that implements the edit command. When I was linking in the autocomplete, I couldn't get it working well with relative paths, but I could use absolute. So instead of taking a relative path within the `rendered` directory, the user can tab complete `/var/lib/metalware/rendered` and then select the file from there.

It does mean the edit command can open any file in the filesystem, however this shouldn't be a big deal. Currently it uses the editor defined by VISUAL first or EDITOR as a fallback. If both haven't been set, then `vi` is used. Possible add a message about using `vi`?

### DIFF
--- a/src/commands/edit.rb
+++ b/src/commands/edit.rb
@@ -24,16 +24,17 @@
 
 require 'system_command'
 require 'metal_log'
+require 'file_path'
 
 module Metalware
   module Commands
     class Edit < CommandHelpers::BaseCommand
       private
 
-      attr_reader :file
+      attr_reader :relative_file_path
 
       def setup
-        @file = args[0]
+        @relative_file_path = args[0]
       end
 
       def run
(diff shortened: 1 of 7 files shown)


src/commands/edit.rb
```diff
@@ -0,0 +1,64 @@
...
+          system_visual
+        elsif !system_editor.empty?
+          system_editor
+        else
+          'vi'
+        end
+      end
+
+      def system_visual
+        @system_visual ||= SystemCommand.run('echo $VISUAL').chomp
```
bobwhitelock: A simpler way to handle this and `system_editor` would be to use `ENV` I think, e.g. `ENV['VISUAL']`.

WilliamMcCumstie:
Simplified the logic and implemented relative paths

PR #174, opened 2017-08-23
Restart named after each update

WilliamMcCumstie:
Fixes the issue logged in #108 

Named needs to be restarted when ever an update command is ran, not
only when it is started.

The methods have been slightly renamed. During setup, start_named is
called to run the bash commands which end with restart_named. But if
named, has already been setup, then restart_named is called only.

The test have been updated to reflect this

### DIFF
--- a/spec/dns/named_spec.rb
+++ b/spec/dns/named_spec.rb
@@ -52,6 +52,7 @@ RSpec.describe Metalware::DNS::Named do
       allow(named).to receive(:setup?).and_return(false)
       # Prevents restart_named from running as the bash commands can't be mocked
       allow(named).to receive(:restart_named)
+      allow(named).to receive(:start_named)
     end
 
     it "errors if external dns isn't set" do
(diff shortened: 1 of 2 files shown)


spec/dns/named_spec.rb
```diff
@@ -74,7 +75,8 @@
     before :each do
       allow(named).to receive(:setup?).and_return(true)
       # Prevents restart_named from accidentally running, DANGEROUS
```
bobwhitelock: Should this comment be changed refer to `start_named` now? Also, can we change it to mention what is 'DANGEROUS', similar to https://github.com/alces-software/metalware/pull/174/commits/f934d314a5810d669cd7a71c0100d38e2f20af9d#diff-a753d91681a41b66f585bed90c0db93fR53?

src/dns/named.rb
```diff
@@ -96,18 +97,22 @@ def render_repo_named_conf
...
         MetalLog.info 'Restarting named'
-        RESTART_NAMED_CMDS.each do |cmd|
+        START_NAMED_CMDS.each do |cmd|
           SystemCommand.run(cmd)
         end
+        restart_named
+      end
+
+      def restart_named
+        SystemCommand.run('systemctl restart named')
```
bobwhitelock: Not a big deal, but you could factor out this string from here and in `START_NAMED_CMDS` to a new constant, to indicate the same command is run in both cases.

WilliamMcCumstie:
I have made those changes in the latest commit

bobwhitelock:
Thanks

Issue #164, opened 2017-08-15
Multiple modes for pxe booting


ste78:
We need to support UEFI booting, to do that we need to extend the legacy PXE templating components. Primary differences in the process are:

- The template used to generate the actual boot file is grub based, which basically means it looks completely different to the current pxelinux.cfg template so potentially we need the option of 2 defaults (one for legacy and one for uefi)
- rather than being saved to /var/lib/tftpboot/pxelinux.cfg, files would need to be saved to /var/lib/tftpboot/efi/ 
- rather than being named <`gethostid 10.10.1.1`> the files need to be named <grub.cfg-`gethostid 10.10.0.1`> eg `/var/lib/tftpboot/pxelinux.cfg/0A0A0101` becomes `/var/lib/tftpboot/efi/grub.cfg-0A0A0101`
- the dhcp host entry needs to be slightly different on one of the fields, this could be done with the default dhcp template if we had either a config value or alces magic value to select the boot type of the node. 

IRL Nodes can only support legacy booting or UEFI booting at any one time so thinking about how this could be implemented, we could possibly have a config value for a node that said if it supports legacy or UEFI booting, defaulting to legacy booting at the domain level.

bobwhitelock:
We implemented the ability to have different `build_method`s the other day, to allow a `basic` `build_method` along with the default (and previously only) `kickstart` `build_method`. From an initial look at this issue it seems the most straightforward way to add this would be to add a new build method to the existing 2, called for example `uefi`, which appropriately renders the `kickstart` along with a new `uefi` template. The `dhcp` template could then be changed appropriately to switch on the build method.

How does this sound @ste78?

ste78:
yeah would probably do it, but it's still primarily a kickstart build method, so maybe call it `uefikickstart` or similar

PR #176, opened 2017-08-25
Allow nodes to build using UEFI

WilliamMcCumstie:
As `UEFI` build is similar to the `kickstart` build, `BuildMethods::UEFI` is inherited from `Kickstart`.

The build is specified by a build type: `uefi` and uses a default template `pxelinux/uefi/default`. This default is still to be added to the repo. To override the template, a config parameter overriding `pxelinux/uefi` is required explicitly. Overriding `pxelinux` will not alter the `uefi` default.

Talking with @ColonelPanicks, the changes required to `DHCP` is can be handled by the repo and thus are not included in this PR.

fixes: #164


WilliamMcCumstie:
The default template location for `uefi` build has been moved. Instead if will look for the template at `<repo>/uefi/default`. Also, in the config the template is not specified under the `pxelinux/uefi` key anymore. Instead it is specified under the `uefi` key.

WilliamMcCumstie:
Removed the setup methods from `named`. Fixes #183

spec/build_methods/uefi_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+  let :build_uefi { Metalware::BuildMethods::UEFI.new(config, node) }
+  let :file_path { Metalware::FilePath.new(config) }
+  let :filesystem do
+    FileSystem.setup(&:with_minimal_repo)
+  end
+
+  it 'determines the correct UEFI pxelinux template' do
+    filesystem.test do
+      path = build_uefi.send(:template_path, :'pxelinux/uefi', node: node)
```
bobwhitelock: From your comment at https://github.com/alces-software/metalware/pull/176#issuecomment-325696451 I gathered we'd changed the UEFI PXELINUX template to be stored at `/uefi`, in which case this is incorrect. This makes me notice that this test is effectively useless, as we're just testing the `BuildMethod#template_path` method which is adequately tested elsewhere, so we can remove this test I think.

src/node.rb
```diff
@@ -185,6 +185,8 @@ def build_method
 
     def build_method_class
       case repo_config[:build_method]&.to_sym
+      when :uefi
```
bobwhitelock: From https://github.com/alces-software/metalware/issues/164#issuecomment-322753763 Steve says this is primarily a Kickstart build method. Unless you've heard differently since then I'd probably call this `uefi-kickstart`.

spec/build_methods/uefi_spec.rb
WilliamMcCumstie: Fixed in: 0ab9c06e38f00beb9337d9884171117346b0dd9d

src/node.rb
WilliamMcCumstie: `uefi` has been renamed in https://github.com/alces-software/metalware/pull/176/commits/67406f0b158d5702f3ce62c7b5cc8d18c1412ecd

WilliamMcCumstie:
I have cherry picked from https://github.com/alces-software/metalware/pull/176/commits/10df831d92b0e8dc123dd4bf1d44a540b28281bb as I needed it own this branch. This should hopefully prevent a conflict when merging. It allows an entire directory to be loaded as is useful for changing the directory structure.

Otherwise the additional code allows for the sharing of as much code as possible between the kickstart methods. The `TEMPLATES` constant is set to include the `REPO_DIR` which is either `pxelinux` or `uefi-kickstart`. This way we do not need to remember to change it in multiple locations

bobwhitelock:
Looks great, thanks!

PR #178, opened 2017-08-30
Fixes Validation::Configure ignore failures bug

WilliamMcCumstie:
Previously the `validate` method on `Validation::Configure` would return an object which contains if the validation succeeded or not. However this was not the object that `success?` used. Instead a stale instance variable was used and thus the validation could fail but still return true. This PR fixes the bug by saving the return object of `validate` as the instance variable.

Also added additional validation checks to allow empty strings as a valid default and to allow `Boolean` default values. Previously these caused the validation to fail but it wasn't being detected due to the previously mentioned bug.

### DIFF
--- a/spec/validation/configure_spec.rb
+++ b/spec/validation/configure_spec.rb
@@ -43,6 +43,11 @@ RSpec.describe Metalware::Validation::Configure do
           type: 'integer',
           default: 10,
         },
+        boolean_true: {
+          question: 'Can I have a boolean true (/yes) default?',
+          type: 'boolean',
+          default: 'yes',
+        }
       },
 
       group: {
@@ -54,6 +59,11 @@ RSpec.describe Metalware::Validation::Configure do
           question: 'Am I a integer question without a default?',
           type: 'integer',
         },
+        boolean_false: {
+          question: 'Can I have a boolean false (/no) default?',
+          type: 'boolean',
+          default: 'no',
+        }
       },
 
       node: {
@@ -236,6 +246,19 @@ RSpec.describe Metalware::Validation::Configure do
                                     },
                                   })
       expect(run_configure_validation(h).keys).to eq([:default_integer_type])
+    end
+  end
+
+  context 'with invalid boolean questions' do
+    it 'fails with non-boolean default' do
+      h = correct_hash.deep_merge(node: {
+                                    bad_integer_question: {
+                                      question: 'Do I fail because my default is a string?',
+                                      type: 'boolean',
+                                      default: 'I am not valid',
+                                    },
+                                  })
+      expect(run_configure_validation(h).keys).to eq([:default_boolean_type])
     end
   end
 end
--- a/spec/validation/configure_spec.rb
+++ b/spec/validation/configure_spec.rb
@@ -62,6 +62,10 @@ RSpec.describe Metalware::Validation::Configure do
           type: 'string',
           default: 'yes I am a string',
         },
+        string_empty_default: {
+          question: 'My default is a empty string?',
+          default: '',
+        },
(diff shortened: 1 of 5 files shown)


spec/validation/configure_spec.rb
```diff
@@ -43,6 +43,11 @@
           type: 'integer',
           default: 10,
         },
+        boolean_true: {
+          question: 'Can I have a boolean true (/yes) default?',
+          type: 'boolean',
+          default: 'yes',
```
bobwhitelock: I don't think we want to require `boolean` questions have their default specified as a `'yes'` or `'no'` string. Firstly this conversion is already performed within the `Configurator`, so requiring this here is unnecessary and also won't work once the default value is passed to the `Configurator` I believe.

Secondly, the default should be the default value which will be saved if no input is given, via whatever UI the user is using to perform the `configure`; in the case of `boolean` questions this will be `true`/`false` rather than `'yes'`/`'no'`. Doing this makes clear what this value is, and also means we can handle a default from a higher level saved answer file in the same way as a `configure.yaml`-defined default.

The use of `'yes'`/`'no'` to enter a default on the CLI can be considered a detail of this particular interface, used because accepting text is the simplest/standard way to get user input in a CLI. Other interfaces could accept different input to set a `boolean` value however, e.g. the GUI could use a check box.

Can we change this validation to validate for `true`/`false`?

src/validation/configure.rb
```diff
@@ -41,31 +41,32 @@ def initialize(file)
...
           end
+          configure_results
         end
-        @configure_results
       end
 
       def success?
-        validate if @configure_results.nil?
-        @configure_results.success?
+        validate.success?
```
bobwhitelock: You could remove this method entirely and replace it with `delegate :success?, to: :validate` at the top of this file. You may or may not need to add `require 'active_support/core_ext/module/delegation'` to ensure `delegate` is imported.

spec/validation/configure_spec.rb
```diff
@@ -247,6 +247,17 @@ def run_configure_validation(my_hash = {})
                                   })
       expect(run_configure_validation(h).keys).to eq([:default_integer_type])
     end
+
+    it 'fails if the default is an empty string' do
```
bobwhitelock: I don't follow why we need a special case for empty string defaults - can't they be covered by the same validating as non-empty strings? Then as I understand it we wouldn't need this test or special handling for empty strings, as this is just another test for `default_type?`. Am I missing something? This validations stuff still looks like magic to me.

src/validation/configure.rb
WilliamMcCumstie: Fixed in: https://github.com/alces-software/metalware/pull/178/commits/f40f29e5897301f20092cb1fe8f4c7afb55978ea

spec/validation/configure_spec.rb
WilliamMcCumstie: Because of the crazy logic that occurs in the `Validation` I can't guarantee that empty strings will be treated the same way as a regular string. The new `Validator` class in a latter PR handles it better.

spec/validation/configure_spec.rb
WilliamMcCumstie: Will look into this. However it isn't ideal to fix it on this Pr but instead fix the issue off #185

WilliamMcCumstie:
Fixed/ responded to your comments. Some of them might need to be fixed in a latter PR.

This PR has been mostly superseded, but it is the foundation of the refactored code.

bobwhitelock:
Thanks, so long as those 2 comments are noted to be looked at then I'll merge this now.

PR #180, opened 2017-08-30
Create the `configure self` command

WilliamMcCumstie:
Based on the updated Validation. Please merge #178 first.

This is the first step for configuring the node metalware runs on. This will be defined as the `self` node. Currently the command runs the standard domain level template rerender.

However it does not add the `self` node to `/etc/hosts` or the genders file. I am not sure if this will be required so it has not been included in this PR. Future PR's will allow the building of the `self` node.

### DIFF
--- a/src/cli_helper/config.yaml
+++ b/src/cli_helper/config.yaml
@@ -69,6 +69,15 @@ subcommands:
       templates themselves.
     action: Commands::Configure::Rerender
 
+  configure_self: &configure_self
+    syntax: metal configure self [options]
+    summary: Configure the master Metalware node
+    description: >
+      Configures the node that metalware is running on. This is treated as the
+      metalware master node. This must be ran to configure metalware to match
+      the system environment. This command is intended to only be ran once.
+    action: Commands::Configure::Self
+
   remove_group: &remove_group
     syntax: metal remove group GROUP_NAME [options]
     summary: Removes a previously configured group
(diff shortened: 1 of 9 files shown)


src/cli_helper/config.yaml
```diff
@@ -56,6 +56,15 @@ subcommands:
       templates themselves.
     action: Commands::Configure::Rerender
 
+  configure_self: &configure_self
+    syntax: metal configure self [options]
+    summary: Configure the master Metalware node
+    description: >
+      Configures the node that metalware is running on. This is treated as the
+      metalware master node. This must be ran to configure metalware to match
+      the system environment. This command is intended to only be ran once.
```
bobwhitelock: > This command is intended to only be ran once.

Is there any negative effect if it is run more than once?

WilliamMcCumstie: Probably, their are a whole bunch of bash scripts that run which probably can't be reran. But it is a repo, kb issue.

bobwhitelock: OK - I guess we don't need to do anything special then to stop someone shooting themselves in the foot due to a bad repo, especially since this is just one of many ways they could do this. This LGTM then once #178 is merged.

WilliamMcCumstie:
Rebased on develop

bobwhitelock:
@WilliamMcCumstie: can this be merged or does anything else need to happen?

WilliamMcCumstie:
Ready to be merged

PR #181, opened 2017-08-31
Support the `build self` command

### LINKED ISSUE #141: Metal host self configuration
From a clean CentOS image we want to be able to run the metalware installer, pull the default repo, run a 'metal configure domain' and the metalware host should be configured to be a deployment server with configuration specific to the domain.

A basic list of the information that would affect how the metal host was configured would be:
- PRV/MGT/EXT IP's and existence of those networks / masks gateways etc. 
- DOMAINNAMES for each network
- External service information (dns servers / proxies / ntp servers etc)

After that information is requested a host configure script would be run, example of how that would look is here: https://github.com/alces-software/knowledgebase/blob/master/epel/7/metalware/metalware.sh

That should be re-runable so every time 'metal configure domain' is run the host will reconfigure for any changes to network details etc. 

Having this allows us to then be able to ship part configured metalware appliances, on which the first thing that happens is a reconfiguration to suit the environment.

WilliamMcCumstie:
This PR is based of #180. Please merge it first.
Fixes #141 

The `build` command now supports the `self` build type. This `self` build operates the same manner as a regular `basic` node build with a few exceptions. 

When building the `self` node, only the self `build_type` can be used. If the `build_type` is not included, `self` will be implicitly used (without a warning). However if `build_type` is set to anything other than `self`, it will result in an error. Similarly if a regular node tries to use the `self` build type, this will also result in an error.
**tl;dr** Self nodes must use the self build_type. Regular nodes can not use the self build_type

The self build_type, uses the `repo/self/default` template. This can be override in `template` section in the config same as any other template using the `self` key.

This PR also contains a commit (8c200ff), for dynamically requiring a directory. This is now being used in a few places, so the code has been included as a utility. It will require any file in a directory relative to the calling file.

### DIFF
--- a/spec/templating/configuration_spec.rb
+++ b/spec/templating/configuration_spec.rb
@@ -1,0 +1,53 @@
+
+# frozen_string_literal: true
+
+#==============================================================================
+# Copyright (C) 2017 Stephen F. Norledge and Alces Software Ltd.
+#
+# This file/package is part of Alces Metalware.
+#
+# Alces Metalware is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License
+# as published by the Free Software Foundation, either version 3 of
+# the License, or (at your option) any later version.
+#
+# Alces Metalware is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this package.  If not, see <http://www.gnu.org/licenses/>.
+#
+# For more information on the Alces Metalware, please visit:
+# https://github.com/alces-software/metalware
+#==============================================================================
+
+require 'templating/configuration'
+require 'config'
+require 'node'
+
+RSpec.describe Metalware::Templating::Configuration do
+  let :config { Metalware::Config.new }
+  let :testnode do
+    double('Metalware::Node',
+           groups: ['group1', 'group2'],
+           name: 'testnode')
+  end
+
+  def configuration_node(node_name)
+    Metalware::Templating::Configuration.for_node(node_name, config: config)
+  end
+
+  describe '#configs' do
+    it 'returns domain and self for the metalware master node' do
+      expect(configuration_node('self').configs).to eq(['domain', 'self'])
+    end
+
+    it 'returns the configs for a regular node' do
+      allow(Metalware::Node).to receive(:new).and_return(testnode)
+      expected_configs = ['domain', 'group2', 'group1', testnode.name]
+      expect(configuration_node(testnode.name).configs).to eq(expected_configs)
+    end
+  end
+end
--- a/spec/templating/configuration_spec.rb
+++ b/spec/templating/configuration_spec.rb
@@ -35,19 +35,20 @@ RSpec.describe Metalware::Templating::Configuration do
            name: 'testnode')
   end
 
-  def configuration_node(node_name)
+  def node_config(node_name)
     Metalware::Templating::Configuration.for_node(node_name, config: config)
+                                        .configs
   end
 
   describe '#configs' do
     it 'returns domain and self for the metalware master node' do
-      expect(configuration_node('self').configs).to eq(['domain', 'self'])
+      expect(node_config('self')).to eq(['domain', 'self'])
     end
 
     it 'returns the configs for a regular node' do
       allow(Metalware::Node).to receive(:new).and_return(testnode)
       expected_configs = ['domain', 'group2', 'group1', testnode.name]
-      expect(configuration_node(testnode.name).configs).to eq(expected_configs)
+      expect(node_config(testnode.name)).to eq(expected_configs)
     end
   end
 end
--- a/src/build_methods/self.rb
+++ b/src/build_methods/self.rb
@@ -1,0 +1,18 @@
+
+# frozen_string_literal: true
+
+module Metalware
+  module BuildMethods
+    class Self < BuildMethod
+      TEMPLATES = [:self].freeze
+
+      def render_build_started_templates(parameters)
+        render_template(:self, parameters: parameters)
+      end
+
+      def render_build_complete_templates(_parameters); end
+
+      private
+    end
+  end
+end
--- a/src/build_methods/self.rb
+++ b/src/build_methods/self.rb
@@ -1,6 +1,8 @@
 
 # frozen_string_literal: true
 
+require 'system_command'
+
 module Metalware
   module BuildMethods
     class Self < BuildMethod
@@ -8,6 +10,11 @@ module Metalware
 
       def render_build_started_templates(parameters)
         render_template(:self, parameters: parameters)
+      end
+
+      def start_build
+        rendered_self_template = file_path.template_save_path(:self, node: node)
+        puts SystemCommand.run("bash #{rendered_self_template}")
       end
 
       def render_build_complete_templates(_parameters); end
--- a/src/build_methods/self.rb
+++ b/src/build_methods/self.rb
@@ -11,8 +11,6 @@ module Metalware
       end
 
       def render_build_complete_templates(_parameters); end
-
-      private
     end
   end
 end
--- a/src/build_methods/self.rb
+++ b/src/build_methods/self.rb
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 require 'system_command'
+require 'fileutils'
 
 module Metalware
   module BuildMethods
@@ -14,7 +15,8 @@ module Metalware
 
       def start_build
         rendered_self_template = file_path.template_save_path(:self, node: node)
-        puts SystemCommand.run("bash #{rendered_self_template}")
+        FileUtils.chmod 'u+x', rendered_self_template
+        puts SystemCommand.run(rendered_self_template)
       end
 
       def render_build_complete_templates(_parameters); end
--- a/src/commands/build.rb
+++ b/src/commands/build.rb
@@ -60,6 +60,7 @@ module Metalware
           puts(EDIT_START_MSG)
           return
         end
+        start_build
         wait_for_nodes_to_build
         teardown
       rescue
@@ -112,6 +113,10 @@ module Metalware
             Templater.render_to_file(config, file[:template_path], render_path, parameters)
           end
         end
+      end
+
+      def start_build
+        nodes.each(&:start_build)
       end
 
       def wait_for_nodes_to_build
--- a/src/commands/build.rb
+++ b/src/commands/build.rb
@@ -116,7 +116,17 @@ module Metalware
       end
 
       def start_build
-        nodes.each(&:start_build)
+        nodes.each do |node|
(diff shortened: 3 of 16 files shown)


ColonelPanicks:
@WilliamMcCumstie it doesn't look to me like the answers for self are being used, see following:
```
# Snippet from config/self.yaml
networks:
  pri:
    defined: <%= alces.answers.pri_network_defined %>
    ip: <%= alces.answers.pri_network_ip %>
    short_hostname: <%= alces.answers.pri_network_short_hostname %>

# Configuring self
[root@localhost repo]# metal configure self
Setup primary network (1/3) [yes/no]  |yes|

IP for nodes on the primary network (2/3)  |10.10.0.1|

Hostname for node (3/3)  |controller.<%= networks.pri.domain %>|

warning: Unset template parameter: nodename
warning: Unset template parameter: nodename
warning: Unset template parameter: nodename
warning: The DNS type has not been set. Reverting to default DNS type: 'hosts'. To
prevent this message, please set "dns_type" in your repo configs, "domain.yaml"
file.
warning: Unset template parameter: nodename

# Building self
[root@localhost repo]# metal build self
error: Unset template parameter: pri_network_defined. Use --trace to view backtrace

# Snippet from config/self.yaml (tweaked)
networks:
  pri:
    #defined: <%= alces.answers.pri_network_defined %>
    defined: true
    ip: <%= alces.answers.pri_network_ip %>
    short_hostname: <%= alces.answers.pri_network_short_hostname %>

# Building self
[root@localhost repo]# metal build self
error: Unset template parameter: pri_network_short_hostname. Use --trace to view backtrace
```

WilliamMcCumstie:
@ColonelPanicks, as discussed irl:

Their where a few bugs that have been fixed. The first is the `self` node answer needed to be saved under the `nodes` directory (fa2e2cb).

Also the rendered self template needed to be ran (29dbb19)

ColonelPanicks:
@WilliamMcCumstie The defaults issue still seems to occur, if I don't manually set the short hostname then I get the following when building (setting the hostname during `metal configure self` allows the build to run through and complete successfully):
```
[root@controller ~]# metal build self --trace
/opt/metalware/src/templating/missing_parameter_wrapper.rb:36:in `method_missing': Unset template parameter: pri_network_short_hostname (Metalware::MissingParameterError)
	from (erb):1:in `block in replace_erb'
	from /opt/metalware/opt/ruby/lib/ruby/2.4.0/erb.rb:893:in `eval'
	from /opt/metalware/opt/ruby/lib/ruby/2.4.0/erb.rb:893:in `block in result'
	from /opt/metalware/opt/ruby/lib/ruby/2.4.0/erb.rb:894:in `result'
	from /opt/metalware/src/templating/renderer.rb:56:in `render_erb_template'
	from /opt/metalware/src/templating/renderer.rb:34:in `replace_erb'
	from /opt/metalware/src/templating/repo_config_parser.rb:132:in `parse_string_config_value'
	from /opt/metalware/src/templating/repo_config_parser.rb:120:in `parse_config_value'
	from /opt/metalware/src/templating/repo_config_parser.rb:123:in `block in parse_config_value'
	from /opt/metalware/src/templating/repo_config_parser.rb:122:in `each'
	from /opt/metalware/src/templating/repo_config_parser.rb:122:in `map'
	from /opt/metalware/src/templating/repo_config_parser.rb:122:in `parse_config_value'
	from /opt/metalware/src/templating/repo_config_parser.rb:123:in `block in parse_config_value'
	from /opt/metalware/src/templating/repo_config_parser.rb:122:in `each'
	from /opt/metalware/src/templating/repo_config_parser.rb:122:in `map'
	from /opt/metalware/src/templating/repo_config_parser.rb:122:in `parse_config_value'
	from /opt/metalware/src/templating/repo_config_parser.rb:113:in `block in perform_config_parsing_pass'
	from /opt/metalware/src/templating/repo_config_parser.rb:112:in `each'
	from /opt/metalware/src/templating/repo_config_parser.rb:112:in `map'
	from /opt/metalware/src/templating/repo_config_parser.rb:112:in `perform_config_parsing_pass'
	from /opt/metalware/src/templating/repo_config_parser.rb:56:in `parse'
	from /opt/metalware/src/templating/repo_config_parser.rb:36:in `parse_for_node'
	from /opt/metalware/src/templater.rb:161:in `initialize'
	from /opt/metalware/src/node.rb:137:in `new'
	from /opt/metalware/src/node.rb:137:in `repo_config'
	from /opt/metalware/src/node.rb:188:in `build_method_class'
	from /opt/metalware/src/node.rb:184:in `build_method'
	from /opt/metalware/src/node.rb:141:in `build_template_paths'
	from /opt/metalware/src/nodes.rb:37:in `each'
	from /opt/metalware/src/nodes.rb:37:in `each'
	from /opt/metalware/src/commands/build.rb:75:in `map'
	from /opt/metalware/src/commands/build.rb:75:in `repo_dependencies'
	from /opt/metalware/src/commands/build.rb:66:in `dependency_hash'
	from /opt/metalware/src/command_helpers/base_command.rb:79:in `enforce_dependency'
	from /opt/metalware/src/command_helpers/base_command.rb:59:in `post_setup'
	from /opt/metalware/src/command_helpers/base_command.rb:39:in `initialize'
	from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/command.rb:217:in `new'
	from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/command.rb:217:in `call'
	from /opt/metalware/src/commander_extensions.rb:61:in `call'
	from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/command.rb:184:in `run'
	from /opt/metalware/src/commander_extensions.rb:49:in `run'
	from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/runner.rb:464:in `run_active_command'
	from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/runner.rb:66:in `run!'
	from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/delegates.rb:15:in `run!'
	from /opt/metalware/src/cli.rb:52:in `run!'
	from /opt/metalware/src/cli.rb:47:in `run'
	from bin/metal:28:in `<main>'
```

WilliamMcCumstie:
@ColonelPanicks,

Looks like the list of default answers that needed to be loaded in was being maintained separately causing the default to be skipped. This has now been fixed by  1aa0097.

src/build_methods/self.rb
```diff
@@ -0,0 +1,23 @@
...
+    class Self < BuildMethod
+      TEMPLATES = [:self].freeze
+
+      def render_build_started_templates(parameters)
+        render_template(:self, parameters: parameters)
+      end
+
+      def start_build
+        rendered_self_template = file_path.template_save_path(:self, node: node)
+        puts SystemCommand.run("bash #{rendered_self_template}")
```
bobwhitelock: Rather than running the rendered script through Bash, can we just run the script directly and obey its `#!` line? This currently works fine as our single repo uses a Bash script, but this would allow a repo to e.g. specify a script with `#! /usr/bin/env ruby` and have it run as expected.

I believe the following should be sufficient to do this, but you may want to quickly test this:

```ruby
rendered_self_template = file_path.template_save_path(:self, node: node)
FileUtils.chmod '+x', rendered_self_template
puts SystemCommand.run(rendered_self_template)

```

src/validation/configure.rb
```diff
@@ -118,7 +129,7 @@ def question_type?(value)
 
         # White-lists the keys allowed in the configure.yaml file
         validate(valid_top_level_keys: :yaml) do |yaml|
-          (yaml.keys - [:domain, :group, :node, :questions]).empty?
+          (yaml.keys - [:domain, :self, :group, :node, :questions]).empty?
```
bobwhitelock: Can the new `Constants::CONFIGURE_SECTIONS` (with `:questions` appended) be used here?

src/commands/build.rb
```diff
@@ -108,7 +109,9 @@ def render_build_files(parameters, node)
       end
 
       def start_build
-        nodes.each(&:start_build)
+        nodes.each do |node|
+          Utils::MetalThread.new { node.start_build }
```
bobwhitelock: I'm not sure we need the `MetalThread` class to manage the build threads - possibly we can use [ThreadGroup](http://devdocs.io/ruby~2.4/threadgroup) and achieve the same thing in a more standardized way? E.g. would the following do the same thing (untested):

```ruby

def build_threads
  @build_threads ||= ThreadGroup.new
end

def start_build
  nodes.each do |node|
    build_threads.add(Thread.new { node.start_build })
  end
end

def clear_up_build_threads
  build_threads.list.map(&:kill)
end

def teardown
  clear_up_build_threads
  clear_up_built_node_marker_files
  Output.stderr 'Done.'
end
```

src/validation/configure.rb
WilliamMcCumstie: This has been implemented in a latter PR: https://github.com/alces-software/metalware/pull/185/commits/3119b417c85b922743e01d46e9614be251798ccd

WilliamMcCumstie:
Rebased on develop

src/commands/build.rb
WilliamMcCumstie: This has been implemented by rebasing the the old commits. https://github.com/alces-software/metalware/pull/181/commits/2345c4d2cfcfbe116eb62fdf5cb5fbd37bff63e7

spec/templating/configuration_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+RSpec.describe Metalware::Templating::Configuration do
+  let :config { Metalware::Config.new }
+  let :testnode do
+    double('Metalware::Node',
+           groups: ['group1', 'group2'],
+           name: 'testnode')
+  end
+
+  def configuration_node(node_name)
+    Metalware::Templating::Configuration.for_node(node_name, config: config)
```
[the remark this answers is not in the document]
WilliamMcCumstie: Fixed https://github.com/alces-software/metalware/pull/181/commits/c8adce896bb905069ab62bf208dfa62f55b5c6c6

src/build_methods/self.rb
WilliamMcCumstie: Fixed in: https://github.com/alces-software/metalware/pull/181/commits/722d944c66cd0f3a6dc3041476bf69ce63950be6

WilliamMcCumstie:
Made those fixes you have suggested. The `build_type` fx is in: https://github.com/alces-software/metalware/pull/181/commits/39cfc9287bf1353ba0a44dedf73e9aa2b3d67efe

Otherwise it should be ready to merge. I have rebased so the `SHA` references in the comments are for the original commits.

bobwhitelock:
@WilliamMcCumstie: Thanks for making the changes; this looks good now.

One final point: you overwrote my additions with your new changes; be careful when you use `git push -f`! I'd suggest [always using `git push --force-with-lease` instead](https://developer.atlassian.com/blog/2015/04/force-with-lease/), which will fail if your push would overwrite upstream changes you've not yet pulled. I don't know why this isn't just the behaviour of `--force` really, I can't see why you'd ever want to silently overwrite things you don't know about.

Issue #179, opened 2017-08-30
Type mismatch with dns_type question


ColonelPanicks:
PR with question for metalware-default: alces-software/metalware-default#11

After running a configure on the domain it is not possible to run the command again:
```
[root@controller repo]# metal configure domain
error: Failed to validate answers: /var/lib/metalware/answers/domain.yaml
A type mismatch has been detected in the following question(s):
{:question=>:dns_type, :answer=>"named", :type=>"choice"}
. Use --trace to view backtrace
[root@controller repo]# metal configure domain --trace
/opt/metalware/src/validation/answer.rb:78:in `load': Failed to validate answers: /var/lib/metalware/answers/domain.yaml (Metalware::ValidationFailure)
A type mismatch has been detected in the following question(s):
{:question=>:dns_type, :answer=>"named", :type=>"choice"}
	from /opt/metalware/src/validation/loader.rb:66:in `answer'
	from /opt/metalware/src/validation/loader.rb:47:in `domain_answers'
	from /opt/metalware/src/dependency.rb:76:in `validate_answer_file'
	from /opt/metalware/src/dependency.rb:69:in `validate_dependency_value'
	from /opt/metalware/src/dependency.rb:54:in `block (2 levels) in run_dependencies'
	from /opt/metalware/src/dependency.rb:54:in `each'
	from /opt/metalware/src/dependency.rb:54:in `block in run_dependencies'
	from /opt/metalware/src/dependency.rb:48:in `each'
	from /opt/metalware/src/dependency.rb:48:in `run_dependencies'
	from /opt/metalware/src/dependency.rb:40:in `enforce'
	from /opt/metalware/src/command_helpers/base_command.rb:79:in `enforce_dependency'
	from /opt/metalware/src/command_helpers/base_command.rb:59:in `post_setup'
	from /opt/metalware/src/command_helpers/base_command.rb:39:in `initialize'
	from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/command.rb:217:in `new'
	from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/command.rb:217:in `call'
	from /opt/metalware/src/commander_extensions.rb:61:in `call'
	from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/command.rb:184:in `run'
[...]

PR #185, opened 2017-09-05
Switch data structure for configure questions and improve validation

WilliamMcCumstie:
This PR is based off #180. Please merge that into develop first. Then rebase `refactor/configure` on develop.

This PR is the first in a series to refactor the configurator. A future PR will contain the `Saver` method for answer files and then once the `GUI` is merged, `Configurator` can be refactored to accommodate the update. Note that this PR is being merged into a separate branch. That is because this PR will break `Metalware` until the `Configurator` is refactored.

The major change in this PR is updating the data structure the configure file uses. The updated data structure can be found in `configure.yaml` on the branch listed below. That branch on `metalware-default` needs to be merged in once the refactoring is complete.
https://github.com/alces-software/metalware-default/tree/fix/configure-data-structure

The aim of changing the data structure is to make it easier to validate. This also means the error messages aren't as cryptic as the correspond to the actual saved data structure (as opposed to the once created internal within metalware). This fixes the following issue:
https://trello.com/c/0667dsDH/158-make-validatorconfigure-error-messages-more-descriptive

The `Validation::Configure` class has been extensively refactored to simplify how it works. The intermediate data structure that was previously being validated has been removed. Otherwise it functions in the same way as before. Most of the commits are small and concern implementing the previous validation checks as defined by the tests (with a few tweaks).

Also implemented the `choice` questions. It differs from original intended use that `choice` is not a "type" of question. Instead, a question can have a `choice` key containing an array of choices. These choices must match the question type (set independently) and the default must appear in the choice list.

This does not fully fix the issues with the choice questions, but is the first step at resolving:
https://github.com/alces-software/metalware/issues/179

### DIFF
--- a/spec/validation/configure_spec.rb
+++ b/spec/validation/configure_spec.rb
@@ -172,24 +172,34 @@ RSpec.describe Metalware::Validation::Configure do
   end
 
   context 'with invalid question fields' do
-    it 'fails if the question is missing or empty identifier' do
-      h = correct_hash.deep_merge(self: [{ question: 'I have no identifier' }])
-      expect_validation_failure(h, /is missing/)
-      h = correct_hash.deep_merge(self: [{
-                                    question: 'I have no identifier',
-                                    identifier: '',
-                                  }])
-      expect_validation_failure(h, /must be filled/)
+    context 'with invalid identifier' do
+      it 'fails when missing' do
+        h = correct_hash.deep_merge(self: [{ question: 'I have no identifier' }])
+        expect_validation_failure(h, /is missing/)
+      end
+
+      it 'fails when empty' do
+        h = correct_hash.deep_merge(self: [{
+                                      question: 'I have no identifier',
+                                      identifier: '',
+                                    }])
+        expect_validation_failure(h, /must be filled/)
+      end
     end
 
-    it 'fails if question is missing or empty' do
-      h = correct_hash.deep_merge(self: [{ identifier: 'missing_question' }])
-      expect_validation_failure(h, /is missing/)
-      h = correct_hash.deep_merge(self: [{
-                                    question: '',
-                                    identifier: 'no_question',
-                                  }])
-      expect_validation_failure(h, /must be filled/)
+    context 'with invalid question' do
+      it 'fails when missing' do
+        h = correct_hash.deep_merge(self: [{ identifier: 'missing_question' }])
+        expect_validation_failure(h, /is missing/)
+      end
+
+      it 'fails when empty' do
+        h = correct_hash.deep_merge(self: [{
+                                      question: '',
+                                      identifier: 'no_question',
+                                    }])
+        expect_validation_failure(h, /must be filled/)
+      end
     end
 
     it "fails if type isn't supported" do
--- a/spec/validation/configure_spec.rb
+++ b/spec/validation/configure_spec.rb
@@ -25,253 +25,242 @@ require 'validation/configure'
 require 'data'
 
 RSpec.describe Metalware::Validation::Configure do
-  let :correct_hash do
-    {
-      questions: {
-        questions: 'Are not part of the specification of a correct file.',
-        note: 'However they are very commonly associated with configure files',
-        note2: 'All other top level keys apart from questions, domain, group' \
-               'and node will cause an error.',
-      },
+  # let :correct_hash do
+  #   {
+  #     questions: {
+  #       questions: 'Are not part of the specification of a correct file.',
+  #       note: 'However they are very commonly associated with configure files',
+  #       note2: 'All other top level keys apart from questions, domain, group' \
+  #              'and node will cause an error.',
+  #     },
 
-      domain: {
-        string_question: {
-          question: 'Am I a string question without a default and type?',
-        },
-        integer_question: {
-          question: 'Am I an integer question with a default?',
-          type: 'integer',
-          default: 10,
-        },
-        boolean_true: {
-          question: 'Can I have a boolean true (/yes) default?',
-          type: 'boolean',
-          default: 'yes',
-        },
-      },
+  #     domain: {
+  #       string_question: {
+  #         question: 'Am I a string question without a default and type?',
+  #       },
+  #       integer_question: {
+  #         question: 'Am I an integer question with a default?',
+  #         type: 'integer',
+  #         default: 10,
+  #       },
+  #       boolean_true: {
+  #         question: 'Can I have a boolean true (/yes) default?',
+  #         type: 'boolean',
+  #         default: 'yes',
+  #       },
+  #     },
 
-      group: {
-        string_question: {
-          question: 'Am I a string question without a type but with a default?',
-          default: 'yes I am a string',
-        },
-        integer_question: {
-          question: 'Am I a integer question without a default?',
-          type: 'integer',
-        },
-        boolean_false: {
-          question: 'Can I have a boolean false (/no) default?',
-          type: 'boolean',
-          default: 'no',
-        },
-      },
+  #     group: {
+  #       string_question: {
+  #         question: 'Am I a string question without a type but with a default?',
+  #         default: 'yes I am a string',
+  #       },
+  #       integer_question: {
+  #         question: 'Am I a integer question without a default?',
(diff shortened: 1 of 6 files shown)


WilliamMcCumstie:
The reference above needs to be addressed so `Boolean` questions have a `true` and `false` default type instead of `yes` and `no`

bobwhitelock:
@WilliamMcCumstie: when I try to rebase this on `develop` or `refactor/configurator` (they point to the same commit now) I get conflicts - would you be able to rebase this appropriately so it just contains the new commits for me to look at? Thanks

WilliamMcCumstie:
Re-basing the branch now

WilliamMcCumstie:
Fixed the boolean default as well. This change will need to be reflected in the `Configurator` refactor

spec/validation/configure_spec.rb
```diff
@@ -21,66 +21,89 @@
...
+    it 'loads configure file from the repo' do
+      data = { data: 'I am the configure data' }
+      Metalware::Data.dump(file_path.configure_file, data)
+      v = Metalware::Validation::Configure.new(config)
+      expect(v.send(:raw_data)).to eq(data)
+    end
+  end
+
+  context 'with a hash input' do
+    it 'uses the has as the data input' do
```
bobwhitelock: `has` -> `hash`

bobwhitelock:
@WilliamMcCumstie: All the tests are failing on this branch now, can you investigate? :neutral_face:

spec/validation/configure_spec.rb
```diff
@@ -34,244 +41,264 @@
...
+    end
   end
 
   def run_configure_validation(my_hash = {})
-    validator = build_validator(my_hash)
-    validator.validate.messages
+    Metalware::Validation::Configure.new(config, my_hash).data
+  end
+
+  def expect_configure_error(my_hash, msg_regex)
```
bobwhitelock: A more descriptive name for this might be `expect_validation_failure` - what do you think?

WilliamMcCumstie:
Yeah that is to be expected. I haven't linked it into the `Configurator` and that is what is causing the tests to fail. However I need to refactor the `Configurator` anyway. So the input thing is `validation/configurator` tests are passing and I will fix the other broken tests next week.

In the mean time, this PR can be merged into `refactor/configurator` and once all the tests are passing again, it can be merged into `develop`

spec/validation/configure_spec.rb
```diff
@@ -34,244 +41,264 @@
...
-
-    it 'fails if question is missing a title' do
-      h = correct_hash.deep_merge(domain: {
-                                    missing_title_question: {
-                                      default: 'I am missing my question!',
-                                    },
-                                  })
-      results = run_configure_validation(h)
-      expect(results[:parameters][:question][0]).to eq('is missing')
+    it 'fails if the question is missing or empty identifier' do
```
bobwhitelock: Generally if you need to use 'or' or 'and' in your test description it probably means you're testing multiple things in one test, and if you can break it up you should do. Doing this will make your tests clearer and more obvious why one is failing, since it will only be testing one thing. In this case since they don't share any setup it should be straightforward to split the two assertions out into two tests.
 Hide outdated

spec/validation/configure_spec.rb
```diff
@@ -34,244 +41,264 @@
                'and node will cause an error.',
```
bobwhitelock: Can we just make this a comment rather than having it split up across the value for several questions? This is slightly confusing, particularly as these aren't valid questions, and will also make the text harder to edit than if it was all in one comment. It may be clearest if the value for `questions` is just `{}` since we don't care about the value (I think?).

Also this comment is now out of date with the addition of `self` - could be worth changing it to say that the valid top keys are `questions` + the values in `Metalware::Repo::CONFIGURE_SECTIONS`.

spec/validation/configure_spec.rb
```diff
@@ -34,244 +41,264 @@
...
-    it 'fails if question if the title is empty' do
-      h = correct_hash.deep_merge(node: {
-                                    missing_title_question: {
-                                      question: '',
-                                      default: 'I am missing my question!',
-                                    },
-                                  })
-      results = run_configure_validation(h)
-      expect(results[:parameters][:question][0]).to eq('must be filled')
+    it 'fails if question is missing or empty' do
```
bobwhitelock: As above, can we split the two assertions out into two tests?

spec/validation/configure_spec.rb
```diff
@@ -34,244 +41,264 @@
...
+      h = correct_hash.deep_merge(group: [{
+                                    identifier: 'invalid_optional_flag',
+                                    question: 'Do I have a boolean optional input?',
+                                    optional: 'I should be true or false',
+                                  }])
+      expect_configure_error(h, /must be boolean/)
     end
   end
 
   context 'with missing question blocks' do
```
bobwhitelock: You might be able to de-duplicate these with something like (untested):

```ruby
[:domain, :group, :node, :self].each do |question_type|
  it "fails when #{question_type} is missing" do
    correct_hash.delete(question_type)
    expect_configure_error(correct_hash, /is missing/)
  end
end
```

You may or may not think this is worth doing.

spec/validation/configure_spec.rb
```diff
@@ -34,244 +41,264 @@
...
           type: 'integer',
           default: 10,
         },
-        boolean_true: {
+        {
+          identifier: 'boolean_true',
           question: 'Can I have a boolean true (/yes) default?',
           type: 'boolean',
-          default: 'yes',
+          default: true,
```
bobwhitelock: In the commit message (https://github.com/alces-software/metalware/pull/185/commits/8ff5ded96c5a57c12fe0d5f5734397d9e75d60af) you say:

> The Configurator will need to accomidate it when it is refactored

Unless I'm missing what you mean, I think the Configurator should just work if you pass it a `boolean` question with a Boolean `default`; I believe it currently works in this case. I don't think the preceding change would have worked (requiring `'yes'`/`'no'` defaults), but you've removed that so this should be fine now (probably :slightly_smiling_face: ).

WilliamMcCumstie: Will look into this when I refactor `Configurator`. I am really not sure what will happen. But as far as the validation goes, this is the correct behaviour.

spec/validation/configure_spec.rb
WilliamMcCumstie: fixed in: https://github.com/alces-software/metalware/pull/185/commits/b9d2c4a5610a4499e95c29f8c62ab1ff023d2df9 https://github.com/alces-software/metalware/pull/185/commits/be52f5d7d3cd6efce621c6ad5fa836a1ee2b817a

spec/validation/configure_spec.rb
WilliamMcCumstie: fixed: https://github.com/alces-software/metalware/pull/185/commits/bb4f11c336bedf9e1138f9320e151fcca5e974c4

spec/validation/configure_spec.rb
WilliamMcCumstie: Fixed: https://github.com/alces-software/metalware/pull/185/commits/bb4f11c336bedf9e1138f9320e151fcca5e974c4

spec/validation/configure_spec.rb
WilliamMcCumstie: Fixed: https://github.com/alces-software/metalware/pull/185/commits/f29c53c975f72f50e204930a79cad7d34ce093db

spec/validation/configure_spec.rb
WilliamMcCumstie: Fixed in: https://github.com/alces-software/metalware/pull/185/commits/571722574003bcda5991dde8768e97f85b9452fa

spec/validation/configure_spec.rb
WilliamMcCumstie: Fixed in: https://github.com/alces-software/metalware/pull/185/commits/571722574003bcda5991dde8768e97f85b9452fa

WilliamMcCumstie:
This looks good to merge now after making those changes

PR #186, opened 2017-09-05
Implement Saver class for file validation

WilliamMcCumstie:
It is ideal that we validate answer files before we save them to prevent an error from being saved in the first place.

This PR contains the `Saver` class that broadly operates like the `Loader` class. However it uses some meta programming to insure the first input argument is always the data.

The answer validator had to be updated so it could take a hash directly to validate instead of a file path. Also the shared code between the `Loader` and `Saver` have been extracted to a shared class.

NOTE: This is being merged into the `refactor/configure` branch as it is required by the `Configurator`

### DIFF
--- a/spec/validation/answer_spec.rb
+++ b/spec/validation/answer_spec.rb
@@ -25,15 +25,10 @@ require 'validation/answer'
 require 'data'
 require 'config'
 require 'filesystem'
-require 'file_path'
 
 RSpec.describe Metalware::Validation::Answer do
   let :config do
     Metalware::Config.new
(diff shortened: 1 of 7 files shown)


bobwhitelock:
@WilliamMcCumstie: Similarly to previous PR, can you rebase this on `develop` to resolve the conflicts with your already merged PRs? Thanks

WilliamMcCumstie:
This PR is going to be merged to allow the refactoring of `Configurator` to occur. This will then be reviewed as a release candidate.

PR #188, opened 2017-09-07
Fix inheriting of false boolean answers

bobwhitelock:
See https://github.com/alces-software/metalware/commit/5e06759becf234f87ad967c1b6c41880ed3d5eda for main details.

PR #189, opened 2017-09-07
Add `configure` `--answers` option

bobwhitelock:
This PR adds a `--answers` option to the `configure` commands. Given a JSON mapping of question identifiers to answer values, the given answers will be saved instead of asking the configure questions and saving user input.

This was added to be used by the GUI, but this feature is exposed on the command line firstly because it may be useful, and secondly because we want the GUI features to match those available in the CLI as closely as possible.

Based on https://github.com/alces-software/metalware/pull/188.

### DIFF
--- a/src/command_helpers/configure_command.rb
+++ b/src/command_helpers/configure_command.rb
@@ -46,7 +46,14 @@ module Metalware
       end
 
       def answers
-        JSON.parse(options.answers) if options.answers
+        if options.answers
+          JSON.parse(options.answers)
+        else
+          # The `--answers` option has not been passed; the Configurator will
+          # ask the questions on the command line to get the answers to be
+          # saved.
+          nil
+        end
       end
 
       def handle_interrupt(_e)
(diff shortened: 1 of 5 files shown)


WilliamMcCumstie:
Rebased on develop

src/command_helpers/configure_command.rb
```diff
@@ -40,11 +40,15 @@ class ConfigureCommand < BaseCommand
...
 
       def run
-        configurator.configure
+        configurator.configure(answers)
         custom_configuration
         render_domain_templates
       end
 
+      def answers
+        JSON.parse(options.answers) if options.answers
```
WilliamMcCumstie: Could we change this to explicitly return `nil` using a ternary operator? The `nil` is important behaviour for how `configure` works so making it return explicitly would be helpful.

bobwhitelock:
Thanks for the review; addressed your comment so assume this is now good to merge.

PR #190, opened 2017-09-07
GUI supporting refactoring

bobwhitelock:
This PR contains some refactoring to restructure some parts of the CLI to be more usable from the Metalware GUI. From the CLI perspective this should not contain any behaviour changes, but this makes several functions available in a few places which are needed by the GUI.

Based on https://github.com/alces-software/metalware/pull/189.

### DIFF
--- a/src/configurator.rb
+++ b/src/configurator.rb
@@ -30,6 +30,43 @@ HighLine::Question.prepend Metalware::Patches::HighLine::Questions
 
 module Metalware
   class Configurator
+    class << self
+      def for_domain(file_path:)
+        new(
+          configure_file: file_path.configure_file,
+          questions_section: :domain,
+          answers_file: file_path.domain_answers,
+          higher_level_answer_files: []
+        )
+      end
+
+      def for_group(group_name, file_path:)
+        new(
+          configure_file: file_path.configure_file,
+          questions_section: :group,
+          answers_file: file_path.group_answers(group_name),
+          higher_level_answer_files: [file_path.domain_answers]
+        )
+      end
+
+      # Note: This is slightly inconsistent with `for_group`, as that just
+      # takes a group name and this takes a Node object (as we need to be able
+      # to access the Node's primary group).
+      def for_node(node, file_path:)
+        new(
+          configure_file: file_path.configure_file,
+          questions_section: :node,
+          answers_file: file_path.node_answers(node.name),
+          higher_level_answer_files: [
+            file_path.domain_answers,
+            file_path.group_answers(node.primary_group),
+          ]
+        )
+      end
+    end
+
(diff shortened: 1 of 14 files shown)


WilliamMcCumstie:
Rebased on develop

src/utils.rb
```diff
@@ -14,6 +14,16 @@ def commentify(string, comment_char: '#', line_length: 80)
           .join("\n")
       end
 
+      def run_command(command_class, *args, **options_hash)
+        options = Commander::Command::Options.new
```
WilliamMcCumstie: Now that we are running `Metalware` commands within our `metal` commands, we need to fix the logging. Currently we log what is in `ARGV` so running the second `Metalware` command will result in the first command being logged twice. This could be very confusing in the future.

Their are a few possible fixes:
1. Easy: Add a line of logging in `run_command` saying what the new command is (an except the old command gets logged twice)
1. Better: Store the old value of `ARGV` within `run_command` and switch it to what thew new command would be (even though it technically did get ran from the `cli`)
1. Best: Pass in a special tag into `BaseCommand` which will tell it not to log with `ARGV` but to construct a different log message based on it's inputs.

src/configurator.rb
```diff
@@ -30,6 +30,43 @@
 
 module Metalware
   class Configurator
+    class << self
+      def for_domain(file_path:)
```
WilliamMcCumstie: These methods shouldn't really have `file_path` passed into them. Currently this is fine because only the `file_path` is required, however in the `Configurator` will need the `config` so it can create the file `loader`.

So to prevent future conflicts, the `config` should be passed into the methods and then each method creates it's own `FilePath` object based on the `config`. This will temporarily add some overhead. However this will allow me to refactor `Configurator` (and move `FilePath` inside it) without having to make any changes to the `GUI`.

src/utils.rb
bobwhitelock: > Now that we are running Metalware commands within our metal commands

This is not strictly true; we now run Metalware commands within the GUI, but the GUI itself is not a Metalware command. It's true that this is a bit of an issue though; if I check my recent logs when using the GUI I see various output lines like this:

```
I, [2017-09-07T17:08:45.407241 #7290]  INFO -- : metal 
```

I'm not sure which of your 'better' and 'best' suggestions is better; while it's true that the command didn't get run from the CLI it does still sound useful to have the log contain an accurate record of all Metalware commands run, however this was done.

Since the GUI is still a prototype I don't think this is key to fix right now, so I've logged this in https://trello.com/c/f6ktqiiM so that we come back to it.

bobwhitelock:
I've addressed one comment and logged the other, so assume this is now good to merge.

Issue #202, opened 2017-09-15
Cannot configure domain unless an eth0 interface is present


ColonelPanicks:
Even if `build_interface:` is set to `em1` in `server.yaml` the following error is seen when running configure domain:
```
Aborting rendering domain templates; the rendered server config is invalid:
>>> em1 is not a valid interface; valid interfaces: lo, eth0, eth1, eth2
error: An error occurred rendering: metalware_server_config. Use --trace to view backtrace
```

Trace output:
```
coming soon
```

Notes from discussion
> "I was right all along, it was fucked. Apology retracted" - @ste78

ColonelPanicks:
### Fresh install on system with em* interfaces

The following error occurs at the end of first domain configure
```
# Error
error: '/opt/metalware/libexec/determine-hostip eth0' produced error 'Device "eth0" does not exist.'. Use --trace to view backtrace

# Trace
/opt/metalware/src/system_command.rb:70:in `handle_error': '/opt/metalware/libexec/determine-hostip eth0' produced error 'Device "eth0" does not exist.' (Metalware::SystemCommandError)
	from /opt/metalware/src/system_command.rb:41:in `run'
	from /opt/metalware/src/deployment_server.rb:75:in `ip_on_interface'
	from /opt/metalware/src/deployment_server.rb:29:in `ip'
	from /opt/metalware/src/templating/magic_namespace.rb:108:in `hostip'
	from /opt/metalware/src/templating/missing_parameter_wrapper.rb:34:in `method_missing'
	from /opt/metalware/src/templating/missing_parameter_wrapper.rb:34:in `method_missing'
	from (erb):1:in `wrapper_binding'
	from /opt/metalware/opt/ruby/lib/ruby/2.4.0/erb.rb:893:in `eval'
	from /opt/metalware/opt/ruby/lib/ruby/2.4.0/erb.rb:893:in `block in result'
	from /opt/metalware/opt/ruby/lib/ruby/2.4.0/erb.rb:894:in `result'
	from /opt/metalware/src/templating/renderer.rb:56:in `render_erb_template'
	from /opt/metalware/src/templating/renderer.rb:34:in `replace_erb'
	from /opt/metalware/src/templating/repo_config_parser.rb:132:in `parse_string_config_value'
	from /opt/metalware/src/templating/repo_config_parser.rb:120:in `parse_config_value'
	from /opt/metalware/src/templating/repo_config_parser.rb:113:in `block in perform_config_parsing_pass'
[...]

PR #213, opened 2017-09-25
Always default build interface to valid interface

bobwhitelock:
Rather than defaulting to `eth0`, which may not exist depending on the host setup, default to the first network interface, which should always exist. This should work better as a more robust default build interface.

Afaics this appears to work fine, @ColonelPanicks you may want to test and merge with your real usage.

### DIFF
--- a/src/deployment_server.rb
+++ b/src/deployment_server.rb
@@ -50,13 +50,13 @@ module Metalware
       end
 
       def build_interface
-        # Default to 'eth0' if `build_interface` is not set yet. In practise
-        # this _should_ only occur prior to first render of server config, but
-        # this is needed so a valid interface is used when rendering the server
-        # config itself (the template shouldn't normally depend on values
-        # dependent on the `build_interface`, but if it's unspecified the
-        # `alces` namespace will fail to be created).
-        server_config[:build_interface] || 'eth0'
+        # Default to first network interface if `build_interface` is not set
+        # yet. In practise this _should_ only occur prior to first render of
+        # server config, but this is needed so a valid interface is used when
+        # rendering the server config itself (the template shouldn't normally
+        # depend on values dependent on the `build_interface`, but if it's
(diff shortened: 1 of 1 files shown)


bobwhitelock:
@ColonelPanicks

> I think that server.yaml being rendered in domain scope is counter to the purpose of the file, would it be possible to either:
> 
> render server.yaml in self scope
> move build_interface into self.yaml and grab the value from there
> Option 2 would be preferable!

I'm not quite sure what you mean by this; why is it 'counter to the purpose of the file'? This file existed prior to adding any 'self' functionality, so doesn't really relate to it; if you think there is a better way for things to work now that this new functionality has been added feel free to log an issue for us to look at in future. I've not investigated but not sure if moving this to a different file would work though, for the same reason that this file was initially added (see https://github.com/alces-software/metalware-default/pull/9).

Regarding the second point: I'll reply there.

ColonelPanicks:
@bobwhitelock apologies, I could have made that a lot clearer!

What I'm getting at is that the build interface doesn't seem like domain-level information to me. With the introduction of the 'self' concept it looks to me like the build interface should fall under self. I understand that it needs to be outside of the `config/` files so that it can be loaded in earlier but if it could use self's primary interface it would be more fitting. This could be resolved in metalware-default with `build_interface: <%= alces.self.config.networks.pri.interface %>`.

I've redrafted this a few times but your comment on Issue #202 clears up any misunderstanding I had and from the testing I've just done it looks like the change to what `build_interface` defaults to has fixed that issue too.

Hope that helps clear up my initial thinking and cheers for your explanation!

bobwhitelock:
@ColonelPanicks: sure, makes sense, feel free to make that change in metalware-default if you like and it works :slightly_smiling_face:.

bobwhitelock:
@ColonelPanicks: do you want me to make any changes to this or is it mergeable?

PR #219, opened 2017-09-27
Create HashMerger module to handle configs and answers

WilliamMcCumstie:
One of the main features of the proposed Namespaces is the ability to render config and answers on a as need basis. However inorder to do so, their needs to be a clear way load the yaml data and merge them into two hashes (answers and config).  The ultimate goal is to render the configs and answers in the same way.

Originally the class was going to be called `MergerHash` but was latter made into the `HashMegers` module. The purpose of `HashMerger` is to merge the hashes for a particular node/ groups combination. It does not determine which files are required as this is better suited for the future `Namespaces`. Instead it enforces the order in which all config files are merged.

HashMerger takes an array of groups. It assumes the first group is the primary group and merges in the order according to the array. 

It has separate classes for the `Config` and `Answer` as the only difference between the two are the `load_yaml` method.


WilliamMcCumstie:
Please note. Further work has been completed on the `HashMerger`. Instead of passing in the `alces` object, a templater block is going to be used instead. 

This means the `alces` object will not need to be passed in and then deleted from the hash. Refer to 6090bfafb6a1e2b534ecb14da1e1ef1a2c716e60 for further details

PR #234, opened 2017-10-24
Fix the`view` and `view-answer` commands

WilliamMcCumstie:
The implementation of the new namespace has broken the `view-config` and `view-answer` commands.

The `view-config` command has been changed to `view`. It can now be used to view different parts of the `namespace`. So `view domain` will show the entire domain namespace, where `view nodes.node01.config` will show the config for node01.

This has been implemented with the `AlcesCommand` mixin which handles the cli input. To make the command less verbose, the leading `domain`, `groups`, `nodes`, and `local` can be shortened to any number of letters. Thus `view g.group1.answer` will also work.

The `view-answer` command needed to be updated so it used the `alces` namespace to pull its values. Both the `view` and `view-answer` commands do not render the config/ answers. This can not currently be done reliable because domain level configs reference `alces.node` which will never be in scope. Thus the domain config would otherwise not be view-able.

This PR also has major changes in how `AlcesUtils` work mainly in regard to `mock_node`. `mock_node` now creates a entry in the genders file in the faked file-system. This faked genders file is automatically copied to a temporary file which can then be used by the `nodeattr` commands.

Therefore in-order to mock nodes, the genders file must be loaded into the faked filesystem. A mock command has been added to `FileSystem` to complete this.


WilliamMcCumstie:
Currently all the tests that rely on the real filesystem are broken. This is because some of the mocking around the log file has been removed. This is causing a permissions error in travis as the log path doesn't exist.

This will be fixed in the next PR by forcing all tests to use the faked filesystem

PR #251, opened 2017-12-06
Allow orphan nodes to be configured without a group

WilliamMcCumstie:
Fixes #251 

Previously `metal configure node` would fail if a node does not belong to a group. This is because there is no mechanism to add single nodes to the genders file and thus metalware.

It is now possible to add single nodes to an orphan group. This is done when a node that does not exist is configured with `metal configure node`. It will automatically be added to the orphan group. A warning is issued however as there is no mechanism to remove it from the orphan group if it is later reconfigured.

The `local` group has been replaced by the `orphan` group. This means it has a group index of 0 and the `local` node will belong to it. The updated genders template (see link below) will automatically detect the local node as an orphan.
https://github.com/alces-software/metalware-default/pull/19/commits/227db0c8d92d4bf5b8cc06a90af8d3ea0b246f5b

The orphan nodes are saved in the group cache for ease of reference when reading the genders file. This means the nodes will not be lost on multiple re-renders of the file without syncing.

Otherwise the remaining commits where internal metalware changes. See commit message for details.

### DIFF
--- a/spec/alces_utils.rb
+++ b/spec/alces_utils.rb
@@ -162,12 +162,6 @@ module AlcesUtils
       metal_config.cli[:strict] = bool
     end
 
-    # TODO: Remove this. It is only used as part of the testing
-    def alces_default_to_domain_scope_off
-      allow(metal_config).to \
-        receive(:alces_default_to_domain_scope).and_return(false)
-    end
-
     def with_blank_config_and_answer(namespace)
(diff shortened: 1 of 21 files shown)


WilliamMcCumstie:
Fixes #251 

Instead of adding the nodes to `domain.yaml`, they are configured with `metal configure node`

PR #265, opened 2017-12-18
Replace existing `console`, `ipmi` and `power` commands with Ruby

vlj91:


src/commands/ipmi.rb
```diff
@@ -0,0 +1,88 @@
...
+          username = node.config.networks.bmc.bmcuser
+          password = node.config.networks.bmc.bmcpassword
+        end
+        # "-U #{username.to_s} -P #{password.to_s}"
+        "-U admin -P #{password}"
+      end
+
+      def render_hostname
+        node = alces.nodes.find_by_name(node_names[0])
+        node.config.networks.bmc.ip
```
WilliamMcCumstie: Do we want a check that the network is defined?

src/commands/ipmi.rb
```diff
@@ -0,0 +1,89 @@
...
+          "-g #{@args[0]}"
+        else
+          "-H #{render_hostname}"
+        end
+      end
+
+      def render_credentials
+        if @options.group
+          username = alces.domain.config.networks.bmc.bmcuser
+          password = alces.domain.config.networks.bmc.bmcpassword
```
bobwhitelock: Is this strictly correct? If we're running the command for a node we will use the node's BMC username and password (sounds reasonable), but if we're running it for a group we will use the whole domain's username and password - but couldn't these be set specifically for some particular group distinct from the whole domain?

bobwhitelock:
Mostly looks good (though I've not run the commands to test it actually works in practise). I've made a couple of tweaks and one question so far, will take another look tomorrow.

src/commands/ipmi.rb
vlj91: Yep, you are correct! I'll change this to look for a groups config instead, thanks :)

vlj91: Amazing, thanks. I can't be certain but I'd say it's quite an edge case - @ColonelPanicks probably has a better idea if that's likely to occur or not. Typically all nodes in a cluster will have the same BMC user/password though.

src/commands/power.rb
```diff
@@ -0,0 +1,65 @@
...
+  module Commands
+    class Power < Ipmi
+      private
+
+      def command
+        hosts = if @options.group
+                  "-g #{@args[0]}"
+                else
+                  "-H #{render_hostname}"
+                end
```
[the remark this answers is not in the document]
bobwhitelock: In fact both here and in `Console` the `command` methods actually look essentially the same as in `Ipmi` - rather than overriding this and having some duplication between these, would it be possible to just override `render_command` and have that contain just this distinct parts?

For instance in this class I think you would need to move `@args[1].to_s` into `render_command` and then switch on this, and you could then remove the argument to `render_command`; in `Console` you would need to have `render_command` return `-e '&' -I lanplus sol #{type}`.

src/commands/power.rb
```diff
@@ -0,0 +1,65 @@
...
+        when 'status'
+          'chassis power status'
+        when 'cycle'
+          'chassis power cycle'
+        when 'reset'
+          'chassis power reset'
+        when 'sensor'
+          'sensor'
+        else
+          raise 'Invalid power command'
```
bobwhitelock: Using `"Invalid power command: #{cmd}"` could be slightly more informative here.

bobwhitelock: Also I think raising `MetalwareError` rather than `StandardError` would be slightly more consistent with the rest of Metalware, and clearly informs us that this an expected error due to some kind of invalid input (which we may need to catch somewhere) rather than unexpected error due to a bug.

src/commands/ipmi.rb
bobwhitelock: I think this would be a good idea. I think it's fine to just access this part of the config for now as we just have one standard repo and we need to depend on its structure to some extent, but adding a check like this in the `setup` method (before we try to use the config) could be useful as [the BMC network will not always be defined for a node](https://github.com/alces-software/metalware-default/blob/develop/config/domain.yaml#L68):

```ruby
unless object.config.networks.bmc.defined
  raise MetalwareError, "BMC network not defined for #{object.name}"
end
```

src/commands/ipmi.rb
```diff
@@ -0,0 +1,93 @@
...
+        end
+      end
+
+      def render_credentials
+        if @options.group
+          username = group.config.networks.bmc.bmcuser
+          password = group.config.networks.bmc.bmcpassword
+        else
+          username = node.config.networks.bmc.bmcuser
+          password = node.config.networks.bmc.bmcpassword
```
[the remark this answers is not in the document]
vlj91: I was looking to do this when I wrote this, but couldn't find the series of words to Google what I was trying to do. Well done on being better than Google.

src/commands/ipmi.rb
```diff
@@ -0,0 +1,79 @@
...
+    class Ipmi < CommandHelpers::BaseCommand
+      private
+
+      prepend CommandHelpers::NodeIdentifier
+
+      def setup; end
+
+      def run
+        SystemCommand.run(command(node.name)) unless options.group
+        if options.group
```
bobwhitelock: Can we have a single `if` rather than needing `unless` and then `if` to switch on the same condition?

src/commands/power.rb
```diff
@@ -20,12 +22,35 @@
...
+        when 'status'
+          'chassis power status'
+        when 'cycle'
+          'chassis power cycle'
+        when 'reset'
+          'chassis power reset'
+        when 'sensor'
+          'sensor'
+        else
+          raise MetalwareError "Invalid power command: #{cmd}"
```
bobwhitelock: Does this work? I don't think it will:

```ruby
[1] pry(main)> raise StandardError "foo"
NoMethodError: undefined method `StandardError' for main:Object
from (pry):1:in `__pry__'
```

I think you're missing a comma after `MetalwareError`.

src/commands/ipmi.rb
vlj91: Fixed!

src/commands/power.rb
vlj91: You are correct, fixed - thanks! :)

src/commands/ipmi.rb
bobwhitelock: Have you seen this @vlj91?

vlj91: I've removed `render_hostname`, it's not really required - and put the BMC network defined check into `render_credentials` instead.

bobwhitelock:
That should do it, thanks :shipit:

PR #276, opened 2018-01-04
Don't attempt to copy `templates` dir that no longer exists. Fixes #267

### LINKED ISSUE #267: Install Checks for Templates Directory
When installing metalware it checks for the non-existent `templates` directory within the temporary metalware directory. The installation continues and completes without issue however it would be better off without the warning.

```
*snip*
         Runtime ... OK

 > Creating initial directory structure
          Create ... \cp: cannot stat ‘/tmp/metalware.uqqPz5Je/templates’: No such file or directory
OK
*snip*
```

vlj91:
I've seen this error far too many times lately, here is a quick fix

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -91,7 +91,7 @@ chmod a+rw /var/lib/metalware/events
 ln -s /etc/hosts /var/lib/metalware/rendered/system/
 
 mkdir -p "${target}"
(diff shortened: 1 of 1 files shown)


bobwhitelock:
Ta vaughan :+1:

PR #277, opened 2018-01-05
KVM orchestration fixes/improvements

vlj91:
This addresses the comments in #270 and includes the certificate generation at metalware install stage

### DIFF
--- a/scripts/lib/install.functions.sh
+++ b/scripts/lib/install.functions.sh
@@ -27,7 +27,7 @@ install_dist() {
 }
 
 install_certs() {
-    export CA_DIR=/opt/alces/ca_setup
+    export CA_DIR=/var/lib/metalware/certs
     mkdir -p $CA_DIR
 
     if [[ ! -f $CA_DIR/cakey.pem ]]; then
--- a/scripts/lib/install.functions.sh
+++ b/scripts/lib/install.functions.sh
@@ -27,10 +27,8 @@ install_dist() {
 }
 
 install_certs() {
-    doing 'Setup'
     export CA_DIR=/opt/alces/ca_setup
     mkdir -p $CA_DIR
-    say_done $?
 
     if [[ ! -f $CA_DIR/cakey.pem ]]; then
       configure_certificate_authority
@@ -46,7 +44,6 @@ install_certs() {
 }
 
 configure_certificate_authority() {
-    doing 'Configuring certificate authority'
     certtool --generate-privkey > $CA_DIR/cakey.pem
     cat << EOF > $CA_DIR/ca.info
 cn = Alces Software
@@ -58,11 +55,9 @@ EOF
              --template $CA_DIR/ca.info \
              --outfile $CA_DIR/cacert.pem
     cp $CA_DIR/cacert.pem /etc/pki/CA/
-    say_done $?
 }
 
 configure_server_authority() {
-    doing 'Configuring server authority'
     certtool --generate-privkey > $CA_DIR/server-key.pem
     cat << EOF > $CA_DIR/server.info
 organization = Alces Software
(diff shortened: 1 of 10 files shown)


bobwhitelock:
Rebased on `develop` to reconcile this with `develop` and @vlj91's other changes :stuck_out_tongue:

scripts/lib/install.functions.sh
```diff
@@ -25,3 +25,71 @@ install_dist() {
     tar -C "${target}" -xzf "${dep_src}/$name.tar.gz"
     say_done $?
 }
+
+install_certs() {
+    export CA_DIR=/opt/alces/ca_setup
+    mkdir -p $CA_DIR
+
+    if [[ ! -f $CA_DIR/cakey.pem ]]; then
```
bobwhitelock: :cake:

src/commands/ipmi.rb
```diff
@@ -36,26 +36,23 @@ class Ipmi < CommandHelpers::BaseCommand
...
 
-      def ipmi
-        if vm?
-          libvirt_run(libvirt_info[:host], node)
-        else
-          puts "#{node}: #{SystemCommand.run(command(node))}"
-        end
+      def ipmi(node)
+        run_vm(node) if vm?(node)
+        run_baremetal(node) unless vm?(node)
```
bobwhitelock: Splitting out methods for the 2 different `run_*` methods looks good, checking the same condition twice looks a bit odd though - if you want terseness then something like this would be fine:

```ruby
vm?(node) ? run_vm(node) : run_baremetal(node)
```

src/commands/ipmi.rb
```diff
@@ -85,16 +82,12 @@ def node_names
...
       end
 
-      def libvirt_info
-        object = options.group ? group : node
-        {
-          host: object.answer.libvirt_host,
-          vm: object.answer.is_vm,
-        }
+      def libvirt_host(node)
+        node.config.libvirt_host
```
bobwhitelock: Don't think you need this method here any more, as the `libvirt_host` is now only used within `Vm`.

src/vm.rb
```diff
@@ -3,70 +3,66 @@
 require 'config'
 require 'libvirt'
 require 'metal_log'
+require 'namespaces/alces'
 
 module Metalware
   class Vm
-    def initialize(libvirt_host, node, *args)
-      @libvirt ||= Libvirt.open("qemu://#{libvirt_host}/system")
+    attr_reader :node
```
bobwhitelock: Doesn't look like you're using this AFAICS, looks like you always access the instance variable directly - `s/@node/node/g` should do it.

src/vm.rb
```diff
@@ -75,19 +71,32 @@ def running?
...
+      when 3
+        'paused'
+      when 4
+        'shutting down'
+      when 5
+        'off'
+      when 6
+        'crashed'
+      when 7
+        'suspended'
```
bobwhitelock: Could be worth adding an `else` which raises here? Just in case a new state appears in the future so we get an error rather than this just returning `nil`.

scripts/lib/install.functions.sh
```diff
@@ -25,3 +25,71 @@ install_dist() {
     tar -C "${target}" -xzf "${dep_src}/$name.tar.gz"
     say_done $?
 }
+
+install_certs() {
+    export CA_DIR=/opt/alces/ca_setup
```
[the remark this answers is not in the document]
vlj91: Agreed - makes sense

src/commands/ipmi.rb
vlj91: TIL this is a thing.. thanks again

src/commands/ipmi.rb
```diff
@@ -42,8 +42,7 @@ def run
       end
 
       def ipmi(node)
-        run_vm(node) if vm?(node)
-        run_baremetal(node) unless vm?(node)
+        vm?(node? ? run_vm(node) : run_baremetal(node)
```
bobwhitelock: looks like a syntax error: `vm?(node` -> `vm?(node)` :slightly_smiling_face:

bobwhitelock: Ignore me, fixed in a future commit :slightly_smiling_face:

bobwhitelock:
Looks great, thanks @vlj91!

Issue #280, opened 2018-01-05
Certificate generation


vlj91:
Following on from https://alces.slack.com/archives/C5FL99R89/p1515169429000077

The first time the `Vm` class is run, we should check that a certificate for a `node`s `libvirt_host` exists in the specified certificates dir (`/var/lib/metalware/certs`). 

If the certificate does not exist, we should generate it, then provide a message detailing where to copy the two generated certificates to on the `libvirt_host`.

vlj91:
Fixed with #281

PR #281, opened 2018-01-08
Certificate generation

vlj91:
Addresses #280 - generates certificates for a `node`s `libvirt_host` the first time `Vm` class is run. These certificates have to be manually copied to the Libvirt host, then the command re-run afterwards.

### DIFF
--- a/src/vm.rb
+++ b/src/vm.rb
@@ -12,7 +12,7 @@ module Metalware
 
     def initialize(node)
       @node = node
-      certs = Certificates.new(node)
+      certs = Certificate.new(node)
       certs.generate && return unless certs.exist?
     end
 
--- a/src/vm.rb
+++ b/src/vm.rb
@@ -65,9 +65,9 @@ module Metalware
 
     CERT_GENERATION_MSG = <<-EOF
     The following certificates have been generated, copy them
-    to the specified remote directory on julius:
+    to the specified remote directory on #{libvirt_host}:
 
-      from (local): #{CERTS_DIR}/julius-{key,cert}.pem
+      from (local): #{CERTS_DIR}/#{libvirt_host}-{key,cert}.pem
        to (remote): /etc/pki/libvirt/{servercert.pem,/private/serverkey.pem}
     EOF
 
--- a/src/vm.rb
+++ b/src/vm.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'config'
+require 'constants'
 require 'libvirt'
 require 'metal_log'
 require 'namespaces/alces'
@@ -10,7 +11,7 @@ module Metalware
     attr_reader :node
 
     def initialize(node)
-      @libvirt ||= Libvirt.open("qemu://#{node.config.libvirt_host}/system")
+      @libvirt ||= certificate? libvirt : generate_certificate
       @node = node
     end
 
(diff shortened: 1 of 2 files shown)


src/vm.rb
```diff
@@ -61,6 +61,61 @@ def destroy
...
+    end
+
+    def certificate_generation_message
+      <<-EOF
+  The following certificates have been generated, copy them
+  to the specified remote directory on #{libvirt_host}:
+
+    from (local): #{CERTS_DIR}/#{libvirt_host}-{key,cert}.pem
+     to (remote): /etc/pki/libvirt/{servercert.pem,/private/serverkey.pem}
+      EOF
```
bobwhitelock: The formatting on this heredoc looks slightly off - typically you would indent the text 1 level (i.e. 2 spaces for us) from the `<<-EOF`, and in this case you might want the `from` and `to` lines to be indented 1 level further, or at least they should line up with each other.

Also, because of the preceding whitespace on each line the output text will appear unnecessarily indented. There's a method in `active_support`, which we already have as a dependency, which will strip off the shortest level of whitespace of all lines. Just replace the first line with `<<-EOF.strip_heredoc` (you will probably also need to add a `require 'active_support/core_ext/string/strip'` up the top so this method is available).

src/vm.rb
```diff
@@ -61,6 +61,61 @@ def destroy
...
+    def libvirt_host
+      node.config.libvirt_host
+    end
+
+    def libvirt
+      Libvirt.open("qemu://#{libvirt_host}/system")
+    end
+
+    def certificate?
+      File.exist?("#{CERTS_DIR}/#{libvirt_host}-key.pem")
```
bobwhitelock: This file path is duplicated here and in `generate_certificate_key` and `generate_server_certificates` below, so it would make sense to split out a `certificate_key_path` method for it to remove this duplication.

While doing that it may be worthwhile to use `File.join` rather than string interpolation to form this file path - I'm not sure if it will make much difference in this case but generally this is a good practice as it:

a. makes the code less dependent on a particular platform separator (in this case `/`);
b. means you'll still get a nice file path if the things to be joined themselves start or end with a separator;
c. makes your intention obvious.

src/vm.rb
```diff
@@ -61,6 +61,61 @@ def destroy
...
+      puts certificate_generation_message
+      exit
+    end
+
+    def generate_certificate_key
+      SystemCommand.run("certtool --generate-privkey > #{CERTS_DIR}/#{libvirt_host}-key.pem")
+    end
+
+    def generate_certificate_info
+      File.open("#{CERTS_DIR}/#{libvirt_host}.info", 'w') do |f|
```
bobwhitelock: As above, this is duplicated here and in `generate_server_certificates` so it would make sense to split out a `certificate_info_path` method for it, and to also use `File.join` while doing that.

src/vm.rb
```diff
@@ -61,6 +61,61 @@ def destroy
...
+        f.puts 'encryption_key'
+        f.puts 'signing_key'
+      end
+    end
+
+    def generate_server_certificates
+      SystemCommand.run("certtool --generate-certificate \
+                        --load-privkey #{CERTS_DIR}/#{libvirt_host}-key.pem \
+                        --load-ca-certificate #{CERTS_DIR}/cacert.pem \
+                        --load-ca-privkey #{CERTS_DIR}/cakey.pem \
```
bobwhitelock: :cake:

src/vm.rb
```diff
@@ -10,8 +10,8 @@ class Vm
     attr_reader :node
 
     def initialize(node)
-      @libvirt ||= Libvirt.open("qemu://#{node.config.libvirt_host}/system")
       @node = node
+      @libvirt ||= certificate? ? libvirt : generate_certificate
```
bobwhitelock: Now there's a `libvirt` method you could move the initializing of `@libvirt` to this rather than here (which will be possible with the change I suggest in the main review comment).

src/vm.rb
vlj91: Should I be using `File.join` everywhere I'm defining file paths?

vlj91: Ah, you've said yes later on.. ignore!

vlj91:
Thanks for reviewing @bobwhitelock - I've gone through your comments and made the amendments as requested 👍

src/vm/certificate.rb
```diff
@@ -0,0 +1,91 @@
...
+      def exist?
+        File.exist?(certificate_key_path)
+      end
+
+      def generate
+        generate_certificate_key
+        generate_certificate_info
+        generate_server_certificates
+        puts certificate_generation_message
+        exit
```
bobwhitelock: Don't think you need this line now, with the addition of the `return` in `Vm` the command should exit anyway if the certs need to be generated, and removing this will make this class more reusable for other uses (e.g. maybe later we'll want a command which generates several certs at once, this line would make this impossible as-is as we would exit immediately after generating the first).

vlj91:
Fixed that, oops! Thanks

bobwhitelock:
Thanks Vaughan!

Issue #283, opened 2018-01-11
[VM Orchestration] Destroy command requires VMs to be running


ColonelPanicks:
The `destroy` command, as I understand, is designed to delete images and undefine the VM configurations on the master. In the event that the VM config is incorrect it is impossible to start the VMs and the VMs have to be manually removed due to the `destroy` restriction on VMs being running.

vlj91:
This is quite a simple fix - I think I had a check in before to only power off the domains if they are already running. Not sure where that check went, but it can easily be added back again. A domain needs to be off in order to undefine it.

PR #288, opened 2018-01-11
KVM orchestration fixes

### LINKED ISSUE #282: Metalware install final message displays incorrect information about certificates
#280 moved some of the certificate generation round, so libvirt host certificates are now generated post-install. The message after install reads:

```
If the Libvirt integration is required, please copy the following certificates
to the Libvirt host(s):

  from (local): /var/lib/metalware/certs/cacert.pem
   to (remote): /etc/pki/CA/cacert.pem (remote)

  from (local): /var/lib/metalware/certs/server-{key,cert}.pem
   to (remote): /etc/pki/libvirt/{servercert.pem,/private/serverkey.pem}
```

The last section (`server-{key,cert}`) etc should be removed - as that message appears later when configuring nodes.

vlj91:
Fix #282 and #283

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -156,8 +156,5 @@ to the Libvirt host(s):
 
   from (local): $CA_DIR/cacert.pem
    to (remote): /etc/pki/CA/cacert.pem (remote)
-
-  from (local): $CA_DIR/server-{key,cert}.pem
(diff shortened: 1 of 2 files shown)


bobwhitelock:
Looks great, thanks @vlj91!

PR #291, opened 2018-01-12
`Validation::Loader` and `Validation::Configure` refactoring

bobwhitelock:
This PR contains some refactoring of these classes to remove some unnecessary stuff and better separate what's handled by each class; a test covering the existing `Loader` behaviour is also included. See commits for full details of each change.

This is all in preparation for having the `Loader` also load configure questions for plugins, which will occur in a future PR; I'm submitting this separately in case you want to look at this PR independently to check I've not overlooked anything while doing this.

### DIFF
--- a/spec/validation/loader_spec.rb
+++ b/spec/validation/loader_spec.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+#==============================================================================
+# Copyright (C) 2017 Stephen F. Norledge and Alces Software Ltd.
+#
+# This file/package is part of Alces Metalware.
+#
+# Alces Metalware is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License
+# as published by the Free Software Foundation, either version 3 of
+# the License, or (at your option) any later version.
+#
+# Alces Metalware is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this package.  If not, see <http://www.gnu.org/licenses/>.
+#
+# For more information on the Alces Metalware, please visit:
+# https://github.com/alces-software/metalware
+#==============================================================================
+
+require 'validation/loader'
+
+RSpec.describe Metalware::Validation::Loader do
+  describe '#configure_data' do
+    let! :config do
+      # Need to create and cache a config before each test as this is expected
+      # by FilePath.
+      Metalware::Config.cache = Metalware::Config.new
+    end
+
+    let :configure_sections do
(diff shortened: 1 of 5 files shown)


src/validation/loader.rb
```diff
@@ -38,10 +38,11 @@ def initialize(metalware_config, cache_configure: false)
       end
 
       def configure_data
-        return @configure_data if @configure_data
-        tree = Validation::Configure.new(config, repo_configure_data).tree
-        @configure_data = tree if cache_configure
-        tree
+        if cache_configure
+          @configure_data ||= configure_data_tree
```
WilliamMcCumstie: I know this is the result of my original code, however what do you think about changing `configure_cache` to just `cache`?

It makes sense that if this is a class that `caches` it is an all or nothing approach. The default should still be not to `cache`. I'm happy for this change to occur in a latter PR if its not suited here.

src/validation/loader.rb
```diff
@@ -31,18 +31,13 @@
...
-        @cache_configure = cache_configure
       end
 
       def configure_data
-        if cache_configure
-          @configure_data ||= configure_data_tree
-        else
-          configure_data_tree
-        end
+        @configure_data ||= configure_data_tree
```
WilliamMcCumstie: Ignore my previous comment about caching, just saw this commit. This seems a reasonable change. My only comment is we now have the `Loader` class implicitly caching somethings but not others.

I'm not sure if this a problem or not, but it might lead to some inconsistencies in the future.

spec/validation/loader_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+#
+# For more information on the Alces Metalware, please visit:
+# https://github.com/alces-software/metalware
+#==============================================================================
+
+require 'validation/loader'
+
+RSpec.describe Metalware::Validation::Loader do
+  describe '#configure_data' do
+    let! :config do
```
WilliamMcCumstie: Not that this is required for this Spec, but you can also use `include AlcesUtils` to set up the `config`. It will also mocks `Alces.new` to always return the same `alces` object which makes testing it easier.

As this spec doesn't need to use `alces`, it is six of one half dozen of the other.

bobwhitelock: Thanks for the info, I'll use `AlcesUtils` if I need to in future tests :slightly_smiling_face:.

src/validation/loader.rb
bobwhitelock: Thanks for mentioning this, it's worth considering. I think it's probably OK to sometimes cache things and sometimes not - it's pretty standard Ruby to use `||=` to cache some things which won't change and might be slightly slow, which loading the configure data could be as it can now involve loading and parsing/validating several files. From the tests still passing we can see that always caching doesn't change the behaviour (at least the behaviour we care about enough to test) - if we later decide we need to cache more things we can see if that breaks anything and handle it appropriately then.

bobwhitelock:
Thanks for the review @WilliamMcCumstie - I'll merge this through now as there's no changes required :shipit:.

Issue #296, opened 2018-01-17
IPMI Command Errors


ColonelPanicks:
When checking the power status with `metal power` it throws out an error that seems to indicate the the node object is being passed to the command instead of `node.name`:
```
[root@controller repo]# metal power node097 status
error: 'ipmitool -H #<Metalware::Namespaces::Node:0x00000002a9ee88, name: node097>.bmc -U 2 -P liugJac1 chassis power status' produced error 'ipmitool: option requires an argument -- 'H'
```

This effects all arguments to the `metal power` command and, I assume, is why `metal status` is also error-ing on the power check.

bobwhitelock:
@vlj91 I believe you changed how `power` works last - can you investigate when you have a moment?

bobwhitelock:
I believe this was fixed in https://github.com/alces-software/metalware/pull/297, please reopen if not.

Issue #295, opened 2018-01-16
`metal status` errors when checking power status on virtual machines


vlj91:
When running `metal status` against a VM or group containing VMs, all of the VMs return `error` status for power

```Node       | Power      | Ping
---------- | ---------- | ----------
vaughannode01 | error      | ok
vaughannode02 | error      | ok
vaughannode03 | error      | ok
vaughannode04 | error      | ok```

vlj91:
>>>1. have some means of having the `puts` which occur within `ipmi`/`power` happen via some indirect means, so when we're running a command from `metal` directly they go to `stdout` but when run via `Utils.run_command` they go somewhere nice for us to use;


[15:43] 
2. redirect `$stdout` temporarily and use `Utils.run_command`, then use the saved `stdout` output

bobwhitelock:
I think you accidentally some formatting @vlj91 `

PR #297, opened 2018-01-17
IPMI command fixes

vlj91:
This addresses #296 - the whole `node` object was being given when it just requires `node.name` in order to correctly run. Also added `-I lanplus` to the rendered command as @ColonelPanicks pointed out this is required.

Tested this on @ColonelPanicks running controller

### DIFF
--- a/src/commands/ipmi.rb
+++ b/src/commands/ipmi.rb
@@ -56,7 +56,7 @@ module Metalware
       end
 
       def command(host)
-        "ipmitool -H #{host}.bmc #{render_credentials} #{render_command}"
+        "ipmitool -H #{host}.bmc -I lanplus #{render_credentials} #{render_command}"
       end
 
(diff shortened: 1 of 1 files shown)


ColonelPanicks:
@vlj91 Fixes the power command but doesn't seem to address the issues with `metal status`:
```
[root@controller ~]# metal power node097 status
node097: Chassis Power is on
[root@controller ~]# metal status node097
Node       | Power      | Ping
---------- | ---------- | ----------
node097    | error      | ok
```

vlj91:
@bobwhitelock - I'd previously opened #295 about this

bobwhitelock:
Yep looks reasonable, and @ColonelPanicks has confirmed that this works so I'll merge it. Would be useful if you look at #295 whenever you get a chance, not super urgent to me though.

PR #299, opened 2018-01-24
Initial plugin access in node namespace

bobwhitelock:
@WilliamMcCumstie Don't know if you want to have a quick look at this to check I'm not doing anything completely the wrong way with all the namespace stuff? There's not much in this PR, this is just the first step for making a `plugins` field containing information on the enabled plugins for each node available in the node namespace.

I'm submitting this now so I can merge some other stuff based on this, and then later submit some smaller PRs adding for you to have a look at adding the main parts of the plugin namespace.

### DIFF
--- a/spec/filesystem.rb
+++ b/spec/filesystem.rb
@@ -50,6 +50,14 @@ class FileSystem
     def enable_plugin(plugin_name)
       Metalware::Plugins.enable!(plugin_name)
     end
+
+    # Perform arbitrary other FileSystem setup.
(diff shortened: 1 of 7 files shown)


bobwhitelock:
@WilliamMcCumstie I'm going to merge this through so I can also merge some branches based on it, but feel free to take a look when you get a chance and I can make any needed changes in a new branch. Would be useful to get your feedback on how things fit into the namespaces.

Issue #308, opened 2018-02-12
Distinguish user input errors from actual errors in Metalware


bobwhitelock:
Only show `Use --trace to view backtrace` for actual errors and not user input errors, to indicate this is the users fault. Will probably involve patching Commander as well as classifying our internal errors into user input errors and actual errors.

bobwhitelock:
Was done in https://github.com/alces-software/metalware/pull/314

PR #314, opened 2018-02-14
Suppress --trace for user errors

WilliamMcCumstie:
This PR sorts the various different errors into `MetalwareError` and `UserMetalwareError`. It then uses the updated version of `Commander` to suppress the `--trace` prompt when a `UserMetalwareError` is raised.

In the processes of sorting the errors, I have removed the error classes that are no longer being used.

NOTE: The gemfile has not been updated yet. Once the new version of commander has been merged, I will add the final commit.

### DIFF
--- a/src/cli.rb
+++ b/src/cli.rb
@@ -43,6 +43,8 @@ module Metalware
       program :version, '2017.2.1'
       program :description, 'Alces tools for the management and configuration of bare metal machines'
 
+      suppress_trace_classes MetalwareError
+
       CliHelper::Parser.new(self).parse_commands
 
       run!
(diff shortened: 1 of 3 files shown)


src/cli.rb
```diff
@@ -43,6 +43,8 @@ def run
       program :version, '2017.2.1'
       program :description, 'Alces tools for the management and configuration of bare metal machines'
 
+      suppress_trace_class UserMetalwareError
```
bobwhitelock: A comment might be helpful here to indicate that this new option comes from our fork, and what this does.

WilliamMcCumstie: Added

WilliamMcCumstie:
Fixes #308

Issue #284, opened 2018-01-11
Simpler Command for Showing Domain/Group Summary


ColonelPanicks:
Currently, the `metal view` command is useful for checking a ton of configuration parameters at once but it would be helpful to have commands that are simpler. For example, a command to dump a few configuration details to give an idea of the cluster configuration in the event of revisiting a cluster for the first time in a few months. Would be great if it can do the following:

- Display cluster general configuration
  - Cluster Name
  - Domain
  - DNS Config

- Display table of group configuration (where group name is a primary group configured by metalware)

| Group Name | Additional Groups | Nodes |
|-|-|-|
| nodes | physical,compute,cluster,all | node[001-100] |
| infra | virtual,service,cluster,all | infra[01-03] |

bobwhitelock:
Call 'metal overview'.

PR #316, opened 2018-02-15
Implement overview command for groups

WilliamMcCumstie:
This is the first step at implementing #284. The new `overview` command currently displays details about the group.

From the issue description it was determined that the overview command will be dependent on group answers and quite possible configs in the future. To prevent the core metalware repo from becoming dependent on a set answer structure, the content to be displayed is stored in a new repo file `overview.yaml`.

This `overview.yaml` file currently should contain two arrays: `headers` and `fields`. The headers unsurprisingly are the table headers. The fields array contains a list of strings to populate the overview table. All fields are rendered against the group and are intended to use `ERB` tags.

In the process of completing the PR, a bug was detected in `AlcesUtils`. Up until now, when ever a node or group was mocked, it would be partially set as in "scope". To make matters worse, both a group and node could be partially in scope. Note the `alces.scope` method itself was never mocked.

This made the behaviour of the mocked `alces` inconsistent with the real version. Until now, this hasn't been a problem in the spec as the `scope` wasn't changing within a single test.

To fix this bug, the mocking of scopes has been completely removed. This did require modifying the other specs which had previously used this behaviour. The majority of commits are reasonable small as each spec needed to be updated manually.

The usage of `AlcesUtils` is still fairly clunky. A more streamlined `dsl` for mocking nodes and groups would be ideal at some point in the future.

### DIFF
--- a/spec/alces_utils.rb
+++ b/spec/alces_utils.rb
@@ -156,6 +156,7 @@ module AlcesUtils
       allow(node).to receive(:hexadecimal_ip).and_return(hex)
     end
 
+    # TODO: Get the new node by reloading the genders file
     def mock_node(name, *genders)
       AlcesUtils.check_and_raise_fakefs_error
       raise_if_node_exists(name)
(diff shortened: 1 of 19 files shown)


WilliamMcCumstie:
An example/future `overview.yaml` file can be found here: https://github.com/alces-software/metalware-default/commit/5764d0c24f6ee3f5efe34f321bc6cbf5dbaff475

PR #317, opened 2018-02-16
Fix/orphan view answers

### LINKED ISSUE #312: `metal view-answers` does not work for `orphan` nodes
To reproduce:

- Create a node `node01` which isn't part of any existing group.

- Then:
    ```bash
    
    [root@controller(dev) metalware]# bin/metal view-answers node node01
    error: Could not locate required answer file: groups/orphan.yaml. Please run 'metal configure group 
    orphan'. Use --trace to view backtrace
    ```

The concept of `orphan` nodes did not exist when `view-answers` was originally added, which is likely why these do not work.

WilliamMcCumstie:
Based on #316, please merge it first. Fairly simple fix #312. See the commit for details:
https://github.com/alces-software/metalware/commit/0bbac62684acaae41d296cbc6b57f92572093d72

### DIFF
--- a/src/dependency.rb
+++ b/src/dependency.rb
@@ -122,6 +122,9 @@ module Metalware
 
       if validate_directory
         Dir.exist?(path)
+      # The orphan group's answer file is always optional
+      elsif (path == FilePath.group_answers('orphan')) && !File.file?(path)
+        true
       elsif File.file?(path)
         block.nil? ? true : !!(yield path)
       else
--- a/spec/dependency_spec.rb
+++ b/spec/dependency_spec.rb
@@ -105,6 +105,17 @@ RSpec.describe Metalware::Dependency do
       end
     end
 
+    # The orphan group does not require an answer file
+    it 'does not validate groups/orphan.yaml' do
+      filesystem.test do
+        expect do
+          enforce_dependencies(
+            configure: ['groups/orphan.yaml']
+          )
+        end.not_to raise_error
+      end
+    end
+
     context 'when answer files exist' do
       before :each do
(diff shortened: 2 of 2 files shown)


bobwhitelock:
Rebased on `develop`.

src/dependency.rb
```diff
@@ -122,6 +122,9 @@ def valid_file?(dep, value, validate_directory = false, &block)
 
       if validate_directory
         Dir.exist?(path)
+      # The orphan group's answer file is always optional
+      elsif (path == FilePath.group_answers('orphan')) && !File.file?(path)
```
bobwhitelock: This line is very dense - I'd suggest breaking it down into descriptively named variables or methods; if you can name these well enough you might be able to remove the comment as well as it will be obvious what's going on just from the code.

More generally, CodeClimate is now complaining about the complexity of this method, and I agree with it. The natural place to split it seems like it would be to at least split out methods for the `case` statement above and this `if` statement, but feel free to split it up however you want.

WilliamMcCumstie: Yeah, I noticed that when I created the PR however I wouldn't put to much effort in fixing this issue per se.

The `Dependency` class as a whole is quite old code and I don't think a lot of it is relevant anymore. We are switching to a model of validation on the fly, so it is starting to become redundant.

Now that we support building non-groups, we can't validate the group answer file exists. It maybe worth going over `Dependency` and stripping back some of the functionality.

Alternatively, we could move the answer specific validation into the `Loader` class (in addition to the validation it already does).

bobwhitelock: :+1:

PR #319, opened 2018-02-22
Add an domain overview section/table

### LINKED ISSUE #284: Simpler Command for Showing Domain/Group Summary

WilliamMcCumstie:
This PR address the issues highlighted in #316 and adds the last few features that fixes #284.

The structure for `overview.yaml` has been updated to use a single array of hashes. This allows the `header` and `value` to be set together and thus prevents issues with length miss-matches. The validation has thus been removed because it will not cause an error anymore.

Instead, if any field is missing, it will be rendered as an empty string. This should make it more flexible to add an removed columns without having to do it all at once.

A domain table has also been added which contains a single row that is rendered in the `domain` scope. Otherwise the table operates in almost the exact same way as the group table with the exception it does not have a name column. To do this, the table generation code has been generalised into a separate class. This makes testing the command easier.

### DIFF
--- a/spec/commands/overview_spec.rb
+++ b/spec/commands/overview_spec.rb
@@ -6,9 +6,24 @@ require 'alces_utils'
 RSpec.describe Metalware::Commands::Overview do
   include AlcesUtils
 
+  let :overview_yaml do
+    {
+      group: [
+        { header: 'heading1', value: static },
+        { header: 'heading2', value: '<%= group.config.key %>' },
+        { header: 'heading3', value: '' },
+        { header: 'missing-value' },
+        { value: 'missing-header' }
+      ]
+    }
+  end
+  let :static { 'static' }
+  let :headers { overview_yaml[:group].map { |h| h[:header] } }
+
   let :config_value { 'config_value' }
 
   AlcesUtils.mock self, :each do
+    Metalware::Data.dump Metalware::FilePath.overview, overview_yaml
     ['group1', 'group2', 'group3'].map do |group|
       config(mock_group(group), key: config_value)
     end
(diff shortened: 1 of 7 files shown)


spec/commands/overview_spec.rb
```diff
@@ -3,73 +3,116 @@
...
+    expect(Metalware::Commands::Overview::Table).to \
+      receive(:new).once.with(*inputs).and_call_original
   end
 
-  context 'with a mismatch between no. headers/bodies in overview.yaml' do
-    before :each do
-      Metalware::Data.dump(Metalware::FilePath.overview,
-                           headers: ['I', 'have', '4', 'headers'],
-                           fields: ['I', 'have', '5', 'body', 'parts'])
+  context 'without overview.yaml' do
```
bobwhitelock: I'm not sure there's much advantage of printing (mostly) empty tables when `overview.yaml` is missing, though I guess just listing the groups is somewhat useful. What about printing a warning up front if `overview.yaml` is missing, to inform the user of what they would need to do to get more output shown, as I can imagine people being confused about why more output is not being shown. Something like:

```
warning: `overview.yaml` missing for repo; falling back to minimal `overview` format.
```

bobwhitelock:
Also: Code Climate's flagged up a few minor things you should address for this as well.

WilliamMcCumstie:
The corresponding repo change can be found at:
https://github.com/alces-software/metalware-default/tree/feature/overview

spec/commands/overview_spec.rb
WilliamMcCumstie: I am of the opinion if we want to start handling if the the file is missing we should completely commit to validating the file. Currently any structural errors in the `yaml` will create a non-user error which should be caught (an re-raised as a user error).

For the time being, the core functionality has been implemented and I do not want to hold back the `2018.1` release any further.

I have added the validation as a nice thing to have that can be included in the next release.
https://trello.com/c/aKduJydC/250-add-validation-to-overviewyaml

bobwhitelock: I think these are not that closely related and could easily be done independently of each other.

Adding something like
```ruby
unless File.exist? FilePath.overview
  logger.warn '`overview.yaml` missing for repo; falling back to minimal `overview` format.'
end
```
seems a pretty straightforward change which would mitigate an issue I can easily see happening with `overview`, where a user switches the repo they are using and then is confused as to why `overview` is not showing much and not indicating why.

I agree that adding validation of this file seems more involved and for less benefit at this stage.

bobwhitelock:
Thanks for addressing the comments, just https://github.com/alces-software/metalware/pull/319#discussion_r170913427 to look at then this can be merged.

spec/commands/overview_spec.rb
WilliamMcCumstie: Added a the warning

PR #320, opened 2018-02-26
Fix accidental lower casing

bobwhitelock:
This appears to have been accidentally lower-cased in 3bffdf3 and was referring to a non-existent `internalerror` variable.


WilliamMcCumstie:
Opps thanks for fixing that up

PR #324, opened 2018-03-02
Fix/ipmi genders

### LINKED ISSUE #323: [2018.1.0-rc3] Secondary genders not working with IPMI power commands
I have an additional genders group that I'd like to check power on, now, it seems that the templating works without issues for the group (`metal template -g storage`) but if I run power I end up with the following:
```[root@controller [cosma7] repo]# nodeattr -c storage
dellmds1,dellnfs1,dellnfs2,delloss01,delloss02,delloss03,delloss04,delloss05,delloss06,delloss07,delloss08,delloss09,delloss10,delloss11,delloss12,delloss13,delloss14,delloss15
[root@controller [cosma7] repo]# metal power -g mds status
dellmds1: Chassis Power is on
[root@controller [cosma7] repo]# metal power -g oss status
delloss01: Chassis Power is off
delloss02: Chassis Power is off
delloss03: Chassis Power is off
delloss04: Chassis Power is off
delloss05: Chassis Power is off
delloss06: Chassis Power is off
delloss07: Chassis Power is off
delloss08: Chassis Power is off
delloss09: Chassis Power is off
delloss10: Chassis Power is off
delloss11: Chassis Power is off
delloss12: Chassis Power is off
delloss13: Chassis Power is off
delloss14: Chassis Power is off
delloss15: Chassis Power is off
[root@controller [cosma7] repo]# metal power -g nfs status
error: 'ipmitool -H dellnfs1.bmc -I lanplus -U admin -P TrebNed7 chassis power status' produced error 'Error: Unable to establish IPMI v2 / RMCP+ session'. Use --trace to view backtrace
[root@controller [cosma7] repo]# metal power -g storage status
error: undefined method `config' for nil:NilClass. Use --trace to view backtrace
[root@controller [cosma7] repo]# metal power -g storage status --trace
/opt/metalware/src/commands/ipmi.rb:74:in `render_credentials': undefined method `config' for nil:NilClass (NoMethodError)
    from /opt/metalware/src/commands/ipmi.rb:65:in `create_ipmitool_command'
    from /opt/metalware/src/commands/power.rb:33:in `ipmi_command'
    from /opt/metalware/src/commands/ipmi.rb:54:in `run_baremetal'
    from /opt/metalware/src/commands/ipmi.rb:45:in `ipmi'
    from /opt/metalware/src/commands/ipmi.rb:39:in `block in run'
    from /opt/metalware/src/commands/ipmi.rb:38:in `each'
    from /opt/metalware/src/commands/ipmi.rb:38:in `run'
    from /opt/metalware/src/command_helpers/base_command.rb:40:in `initialize'
    from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/command.rb:217:in `new'
    from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/command.rb:217:in `call'
    from /opt/metalware/src/commander_extensions.rb:61:in `call'
    from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/command.rb:184:in `run'
    from /opt/metalware/src/commander_extensions.rb:49:in `run'
    from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/runner.rb:464:in `run_active_command'
    from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/runner.rb:66:in `run!'
    from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/delegates.rb:15:in `run!'
    from /opt/metalware/src/cli.rb:53:in `run!'
    from /opt/metalware/src/cli.rb:48:in `run'
    from bin/metal:28:in `<main>'
```
BMC IPs are configured for all the systems (some are down/unconfigured right now but I wouldn't expect the above error if that were to happen)

This is on a system running 2018.1.0-rc3.

WilliamMcCumstie:
Fixes #323 

The reason why `ipmi` (and thus `power`) did not work with secondary gender groups is becomes of how the credentials where being rendered. It was still assuming that the nodes where in a group and would have a `group.config` containing the `bmc` password.

As genders do not have `config`s, this approach is no longer feasible. Instead, the credentials need to be rendered individually for each node. This should allow it to work in all situations.

The remaining commits clean up to code and the spec.

### DIFF
--- a/src/commands/ipmi.rb
+++ b/src/commands/ipmi.rb
@@ -60,10 +60,10 @@ module Metalware
       end
 
       def create_ipmitool_command(node, host:, arguments:)
-        <<~ECMD.squish
+        <<~COMMAND.squish
           ipmitool -H #{host} -I lanplus #{render_credentials(node)}
           #{arguments}
-        ECMD
+        COMMAND
       end
 
       def command_argument
--- a/src/commands/ipmi.rb
+++ b/src/commands/ipmi.rb
@@ -51,18 +51,14 @@ module Metalware
       end
 
       def run_baremetal(node)
-        puts "#{node.name}: #{SystemCommand.run(ipmi_command(node.name))}"
+        puts "#{node.name}: #{SystemCommand.run(ipmi_command(node))}"
       end
 
-      def ipmi_command(node_name)
-        create_ipmitool_command(
-          host: "#{node_name}.bmc",
-          arguments: command_argument
-        )
-      end
-
-      def create_ipmitool_command(host:, arguments:)
-        "ipmitool -H #{host} -I lanplus #{render_credentials} #{arguments}"
+      def ipmi_command(node)
+        <<~EOF.squish
+          ipmitool -H #{node.name}.bmc -I lanplus #{render_credentials}
+          #{command_argument}
+        EOF
       end
 
       def command_argument
--- a/src/commands/ipmi.rb
+++ b/src/commands/ipmi.rb
@@ -56,8 +56,8 @@ module Metalware
 
       def ipmi_command(node)
         <<~EOF.squish
-          ipmitool -H #{node.name}.bmc -I lanplus #{render_credentials}
-          #{command_argument}
+          ipmitool -H #{node.name}.bmc -I lanplus
+          #{render_credentials(node)} #{command_argument}
         EOF
       end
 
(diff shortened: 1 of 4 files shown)


src/commands/ipmi.rb
```diff
@@ -51,28 +51,28 @@ def run_vm(node)
...
-        )
+      def ipmi_command(node)
+        opt = { host: "#{node.name}.bmc", arguments: command_argument }
+        create_ipmitool_command(node, **opt)
       end
 
-      def create_ipmitool_command(host:, arguments:)
-        "ipmitool -H #{host} -I lanplus #{render_credentials} #{arguments}"
+      def create_ipmitool_command(node, host:, arguments:)
+        <<~ECMD.squish
```
bobwhitelock: What does `ECMD` stand for?

WilliamMcCumstie: End command. `rubocop` complains if you use `EOF`

bobwhitelock: That seems worse as it's non-standard but no more descriptive (`EOF` is a standard delimiter for heredocs) - I think you can ignore rubocop in this case; I'll push a change.

PR #327, opened 2018-03-05
Fix wrong host being used for `metal console`

### LINKED ISSUE #326: metal console no longer working
metal console seems to be broken for ipmi nodes, metal power works as expected but metal console seems to be talking to the wrong host (in this example the ipmi hostname passed to -H should be node001.bmc, not node001)

```
[root@controller [cluster] ~]# metal console node001
error: 'ipmitool -H node001 -I lanplus -U admin -P ******* -e '&' sol info' produced error 'Error: Unable to establish IPMI v2 / RMCP+ session'. Use --trace to view backtrace
```

bobwhitelock:
This PR fixes #326 (see 
https://github.com/alces-software/metalware/commit/5f001c04d0e9b617e146089d962e30986de53a50 for the main fix). Some related refactoring enabled by the `ipmi`-descended commands now being more similar is also included.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ watch-rsync:
 		--name 'Metalware' \
 		--pattern '**/*' \
 		--exit \
+		--no-notify \
 		make rsync
 
 # Note: need to become root for Metalware; -t option allows coloured output.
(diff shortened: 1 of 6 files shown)


bobwhitelock:
Thanks for reviewing - not sure what you mean, doesn't look like this change has been made before?

PR #334, opened 2018-03-08
Allow the spec version of `alces` to be reset

WilliamMcCumstie:
This allows the mocked version of alces to be reset in the spec. Typically this shouldn't be needed in most cases as only one version of alces exists. However if the command being tested changes a file used to generate `alces`, then it can't be reloaded from within the `spec`.

This change allows greater control over what is being mocked. Please note, any mocked nodes and groups will still exist as they alter the `genders` and `group_cache` files. However mocked `config` and `answer` will be reset as this is done the `alces` object itself.

### DIFF
--- a/spec/alces_utils.rb
+++ b/spec/alces_utils.rb
@@ -13,12 +13,19 @@ module AlcesUtils
   class << self
     def start(example_group, config: nil)
       example_group.instance_exec do
-        let! :alces do
-          test_alces = Metalware::Namespaces::Alces.new
+        # Cache the first version of alces to be created
+        # This allows it to be mocked during the spec
+        # It can also be reset in the test (see below)
+        before :each do
           allow(Metalware::Namespaces::Alces).to \
-            receive(:new).and_return(test_alces)
-          test_alces
+            receive(:new).and_wrap_original do |m, *a|
+              @spec_alces ? @spec_alces : (@spec_alces = m.call(*a))
+            end
         end
+
+        # `alces` is defined as a method so it can be reset
+        define_method :alces { Metalware::Namespaces::Alces.new }
+        define_method :reset_alces { @spec_alces = nil }
 
         #
         # Mocks nodeattr to use faked genders file
--- a/spec/alces_utils.rb
+++ b/spec/alces_utils.rb
@@ -172,8 +172,6 @@ module AlcesUtils
       raise_if_node_exists(name)
       add_node_to_genders_file(name, *genders)
       Metalware::Namespaces::Node.create(alces, name).tap do |node|
-        with_blank_config_and_answer(node)
-        hexadecimal_ip(node)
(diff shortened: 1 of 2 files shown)


WilliamMcCumstie:
Hold off merging this for a min

WilliamMcCumstie:
Okay this PR should now be ready to review. It also extracts the method for spoofing the `genders` file. This allows the nodeattr to use the faked genders file without having to mock the entire namespace.

spec/alces_utils.rb
```diff
@@ -13,39 +13,24 @@ module AlcesUtils
...
-          test_alces = Metalware::Namespaces::Alces.new
+        # Cache the first version of alces to be created
+        # This allows it to be mocked during the spec
+        # It can also be reset in the test (see below)
+        before :each do
           allow(Metalware::Namespaces::Alces).to \
-            receive(:new).and_return(test_alces)
-          test_alces
+            receive(:new).and_wrap_original do |m, *a|
+              @spec_alces ? @spec_alces : (@spec_alces = m.call(*a))
```
bobwhitelock: Possibly I'm mistaken, but isn't this functionally equivalent to doing `@spec_alces ||= m.call(*a)`?

spec/alces_utils.rb
```diff
@@ -62,6 +47,35 @@ def nodeattr_cmd_trim_f(command)
...
+          AlcesUtils.check_and_raise_fakefs_error
+          path = AlcesUtils.nodeattr_genders_file_path(args[0])
+          cmd = AlcesUtils.nodeattr_cmd_trim_f(args[0])
+          genders_data = File.read(path)
+          tempfile = nil
+          begin
+            FakeFS.without do
+              tempfile = Tempfile.open('mock-genders')
+              tempfile.write(genders_data)
+              tempfile.close
```
bobwhitelock: I believe you may be able to replace this block with `Tempfile.open('mock-genders') { |f| f.write(genders_data) }`, which is a bit simpler.

spec/alces_utils.rb
WilliamMcCumstie: Done

spec/alces_utils.rb
WilliamMcCumstie: In this situation we can not. The `tempfile` must be opened with `FakeFS` off. However in-order to ensure the file is indeed deleted, the file pointer `tempfile` needs to be unlinked after `nodeattr` is called (thus it is within the ensure block).

This means that the `tempfile` variable needs to be defined outside of the entire `being-ensure` block.

In lining it would result in the number of characters wrote to the file being returned NOT the file pointer thus breaking the unlink.

WilliamMcCumstie: tl;dr No because `FakeFS` is a pain in the ass

bobwhitelock: Fair enough :slightly_smiling_face:

PR #335, opened 2018-03-09
Fix/simplify defaults

WilliamMcCumstie:
Based on #334, please merge it first.

This simplifies the merging of the answers hash. The use of multiple hashes to determine the default has never been used in practice and does not have a spec. I believe it also has a bug within it, however it is hard to tell without the `spec` or being used.

Instead only the `domain` level default shall be used. This greatly simplifies how the default is generated. There maybe a few edge cases where certain questions that aren't defined at the domain level can't have defaults. However these can be addressed after `Configurator` is refactored (next PR).

### DIFF
--- a/spec/hash_mergers/answer_spec.rb
+++ b/spec/hash_mergers/answer_spec.rb
@@ -1,0 +1,105 @@
+
+# frozen_string_literal: true
+
+require 'hash_mergers'
+require 'alces_utils'
+require 'data'
+require 'file_path'
+
+RSpec.describe Metalware::HashMergers::Answer do
+  include AlcesUtils
+
+  let :group { AlcesUtils.mock(self) { mock_group('new_group') } }
+  let :node do
+    AlcesUtils.mock(self) { mock_node('new_node', group.name) }
+  end
+
+  let :identifier { :question_identifier }
+  let :questions do
+    {
+      domain: [{
+        identifier: identifier.to_s,
+        question: 'Ask domain question?',
+        default: 'domain-default',
+      }],
+      group: [{
+        identifier: identifier.to_s,
+        question: 'Ask group question?',
+        default: 'group-default', # Should be ignored
+      }],
+      node: [{
+        identifier: identifier.to_s,
+        question: 'Ask node question?',
+        default: 'node-default', # Should be ignored
+      }],
+      local: []
+    }
+  end
+
+  def answers(namespace)
+    case namespace
+    when Metalware::Namespaces::Domain
+      'domain_answer'
+    when Metalware::Namespaces::Group
+      'group_answer'
+    when Metalware::Namespaces::Node
+      'node_answer'
+    else
+      raise 'unexpected error'
+    end
+  end
+
+  before :each do
+    Metalware::Data.dump Metalware::FilePath.configure_file, questions
+  end
+
+  shared_examples 'run contexts with shared' do |spec_group|
+    context 'when loading domain answers' do
+      subject { alces.domain }
+      include_examples spec_group
+    end
+
+    context 'when loading group answers' do
+      subject { group }
+      include_examples spec_group
+    end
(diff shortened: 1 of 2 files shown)


bobwhitelock:

src/hash_mergers/answer.rb
```diff
@@ -9,33 +9,19 @@ class Answer < HashMerger
...
-
-      def default_hash(section)
-        section_default(section).each_with_object({}) do |(key, value), memo|
-          memo[key] = value.default unless value.default.nil?
-        end
+      def domain_answers
+        loader.flattened_configure_section(:domain)
+              .reject { |_k, value| value.default.nil? }
+              .map { |key, value| [key, value.default] }
+              .to_h
```
bobwhitelock: Rather than `map`ing and then `to_h`ing when you want a hash with only the values changed from an existing hash, you can instead use [`transform_values`](http://devdocs.io/ruby~2.4/hash#method-i-transform_values) which will let you do this in a single line.

spec/hash_mergers/answer_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+        identifier => answers(node)
+      )
+    end
+
+    shared_examples 'uses the saved answer' do
+      it 'uses the saved answer' do
+        expect(subject.answer.send(identifier)).to eq(answers(subject))
+      end
+    end
+    include_examples 'run contexts with shared', 'uses the saved answer'
```
bobwhitelock: I find the amount of indirection in this file quite confusing, with shared examples being passed other shared examples which they then include in various places, although I understand why you're doing it. Possibly we can improve it a bit though. A couple of suggestions you could try:

- Rather than have the name of a shared example be passed to `include_examples 'run contexts with shared'`, it would be clearer I think to pass a hash of the answers which should be used for each namespace. I.e. this would be a hash of each namespace to `'domain-default'` for `'without answer files'`, and a hash of each namespace to the answers as defined in `answers` for `'with answer files'`. This would also DRY up the repeated `expect(subject.answer.send(identifier)).to eq(...)` in each of the passed shared examples.

- I don't think `'run contexts with shared'` is a very clear shared example group name, as it uses the language of RSpec rather than the language of what we are actually testing, which would be more descriptive. It would also become inaccurate following the above suggested change. Possibly 'uses answers for namespaces' would then be a more descriptive name?

src/hash_mergers/answer.rb
```diff
@@ -9,33 +9,19 @@ class Answer < HashMerger
...
-            x.push(default_hash(:node))
-          end
-        end
-      end
-
-      def default_hash(section)
-        section_default(section).each_with_object({}) do |(key, value), memo|
-          memo[key] = value.default unless value.default.nil?
-        end
+      def domain_answers
```
bobwhitelock: I'm not sure I follow this method name - possibly I'm mistaken, but aren't these the 'defaults' rather than the `domain_answers`?

src/hash_mergers/answer.rb
WilliamMcCumstie: Thanks for pointing that out. However, this line of code has been removed in a latter commit.

spec/hash_mergers/answer_spec.rb
WilliamMcCumstie: This works for now, I have created a card for it:
https://trello.com/c/BnJaqzqE/251-clean-up-answer-spec

src/hash_mergers/answer.rb
WilliamMcCumstie: Opps good job picking that up. It's been renamed.

spec/hash_mergers/answer_spec.rb
bobwhitelock: Thanks :slightly_smiling_face:

Issue #337, opened 2018-03-09
Power Command on Groups Exits if One Node Fails


ColonelPanicks:
Title says it all, if `metal power -g nodes status` is run it will exit if one node fails to respond. A node failing to respond should not be viewed as a critical command error.

PR #338, opened 2018-03-09
Make `power` (and `ipmi`) fault tolerant

bobwhitelock:
Fixes https://github.com/alces-software/metalware/issues/337.

### DIFF
--- a/spec/commands/power_spec.rb
+++ b/spec/commands/power_spec.rb
@@ -72,15 +72,25 @@ RSpec.describe Metalware::Commands::Power do
         allow(
           Metalware::SystemCommand
         ).to receive(:run)
-          .with(/ipmitool/)
+          .with(/ipmitool -H node01/)
+          .once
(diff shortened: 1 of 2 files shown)


spec/commands/power_spec.rb
```diff
@@ -58,10 +58,30 @@ def run_power(node_identifier, command, **options)
...
+          expect(output).to include("#{name}: output123")
         end
+      end
+
+      it 'does not error when individual ipmi commands error' do
+        allow(
+          Metalware::SystemCommand
+        ).to receive(:run)
+          .with(/ipmitool/)
+          .and_raise(Metalware::SystemCommandError, 'error123')
```
WilliamMcCumstie: Good start for testing the command but could we also ensure the command is ran the correct number of times? Just in case the error is short circuiting the loop.

Also can we only raise the error once for a single node?

bobwhitelock:
Thanks for the review and suggestion, I've added a commit making that change and will merge this once CI passes.

PR #339, opened 2018-03-09
Dev/more cleaning of configurator

WilliamMcCumstie:
Based on #336, please merge it first.

This PR follows the same tack as the last. It mainly focuses on cleaning up the `Configurator` process to the point that `QuestionTree` is almost completely abstracted away. No when the `Configurator` loops through all the questions, the conditional questions are already skipped AND the progress bar has been set.

A bunch of the logic has also been cleaned up. This mostly entailed deleting logic and checking if tests still passed. Turns out basically none of the behaviour concerning how answers should be saved according to the defaults has been tested. This made refactoring it easy as it is already buggy so I wasn't concerned about more bugs. Primarily the focus was to reduce down all logic that hasn't been tested. See https://github.com/alces-software/metalware/pull/339/commits/d8ba4352673034fba263c6083048a10cec1f3979 and https://github.com/alces-software/metalware/pull/339/commits/192cd9e11070118254b0b4d1b61818a58946a554 and https://github.com/alces-software/metalware/pull/339/commits/b0afe6d0eae4810ef8c53ef405d04bb48ed5df3a

This meant that the `old_answers` do not get used by `highline` anymore. This partly makes sense as they should already been merged into the answers returned from the namespace.

The remained of this PR is focused at integrating the `QuestionTree` object into the namespace. The questions are required by `HashMergers::Answer` and the `Configurator`. So it made sense to define the questions on the namespace so they can be used in both locations. The defaults are now generated by a common method. See https://github.com/alces-software/metalware/commit/4756920110eaa32a38ffdcd8f83ddb0c2d659337

A few of the methods on the `loader` have also been moved onto `QuesionTree` as that is the more appropriate location.

### DIFF
--- a/spec/hash_mergers/hash_merger_spec.rb
+++ b/spec/hash_mergers/hash_merger_spec.rb
@@ -25,7 +25,7 @@ RSpec.describe Metalware::HashMergers::HashMerger do
 
   def build_merged_hash(**hash_input)
     hm = Metalware::HashMergers
-    b = Proc.new { |t| t }
+    b = proc { |t| t }
     OpenStruct.new(
       config: hm::Config.new.merge(**hash_input, &b),
       answer: hm::Answer.new(alces).merge(**hash_input, &b)
--- a/spec/hash_mergers/hash_merger_spec.rb
+++ b/spec/hash_mergers/hash_merger_spec.rb
@@ -28,7 +28,7 @@ RSpec.describe Metalware::HashMergers::HashMerger do
     b = Proc.new { |t| t }
     OpenStruct.new(
       config: hm::Config.new.merge(**hash_input, &b),
-      answer: hm::Answer.new.merge(**hash_input, &b)
+      answer: hm::Answer.new(alces).merge(**hash_input, &b)
     )
   end
 
--- a/spec/hash_mergers/hash_merger_spec.rb
+++ b/spec/hash_mergers/hash_merger_spec.rb
@@ -24,9 +24,12 @@ RSpec.describe Metalware::HashMergers::HashMerger do
   end
 
   def build_merged_hash(**hash_input)
-    Metalware::HashMergers.merge(**hash_input) do |template_string|
-      template_string
-    end
+    hm = Metalware::HashMergers
+    b = Proc.new { |t| t }
+    OpenStruct.new(
+      config: hm::Config.new.merge(**hash_input, &b),
+      answer: hm::Answer.new.merge(**hash_input, &b)
+    )
   end
 
   def expect_config_value(my_hash)
--- a/src/configurator/question.rb
+++ b/src/configurator/question.rb
@@ -60,20 +60,15 @@ module Metalware
       end
 
       def default_input
-        type.boolean? ? boolean_default_input : current_answer_value
+        type.boolean? ? boolean_default_input : default
       end
 
       def boolean_default_input
-        return nil if current_answer_value.nil?
+        return nil if default.nil?
 
         # Default for a boolean question which has a previous answer should be
         # set to the input HighLine's `agree` expects, i.e. 'yes' or 'no'.
-        current_answer_value ? 'yes' : 'no'
-      end
-
-      # The answer value this question at this level would currently take.
-      def current_answer_value
-        default
+        default ? 'yes' : 'no'
       end
 
       def ask_boolean_question
--- a/src/configurator/question.rb
+++ b/src/configurator/question.rb
@@ -18,7 +18,7 @@ module Metalware
       end
 
       attr_accessor :default
-      attr_reader :progress_indicator
+      delegate :identifier, to: :question_node
 
       def ask
         ask_method = choices.nil? ? "ask_#{type}_question" : 'ask_choice_question'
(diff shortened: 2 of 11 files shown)


bobwhitelock:

spec/hash_mergers/hash_merger_spec.rb
```diff
@@ -24,9 +24,12 @@
...
 
   def build_merged_hash(**hash_input)
-    Metalware::HashMergers.merge(**hash_input) do |template_string|
-      template_string
-    end
+    hm = Metalware::HashMergers
+    b = proc { |t| t }
+    OpenStruct.new(
+      config: hm::Config.new.merge(**hash_input, &b),
+      answer: hm::Answer.new(alces).merge(**hash_input, &b)
```
bobwhitelock: `b` here just returns the object it is passed, so you can just replace both uses here with `&:itself` (see [`Object.itself`](http://devdocs.io/ruby~2.4/object#method-i-itself)) and remove defining `b`.

src/question_tree.rb
```diff
@@ -45,20 +46,38 @@ def identifiers
       filtered_each.map(&:identifier)
     end
 
+    # START REMAPPING 'content' to variable names
```
bobwhitelock: I don't think we need these markers indicating what this group of different methods is for, it already seems pretty clear to me what these methods do and adding this is just extra noise which needs to be read. If you really want to keep these methods distinct from others you could extract a well-named mixin for them, for which [`ActiveSupport::Concern`](http://devdocs.io/rails~5.1/activesupport/concern) could be used if you wanted, but I don't think that's necessary.

src/question_tree.rb
```diff
@@ -45,20 +46,38 @@ def identifiers
...
+    delegate :choices, :optional, :type, to: :os_content
```
bobwhitelock: Generally it's clearest to keep all the macros for defining things on a class (`delegate`, `alias` etc) up the top of whatever privacy section they're in (i.e. at the top of the whole class for public methods, or up the top of the `private` section for private methods. Otherwise these may get lost/repeated amongst the rest of the normal method definitions.

src/question_tree.rb
```diff
@@ -27,14 +25,17 @@ class QuestionTree < Tree::TreeNode
     def ask_questions
       filtered_each.with_index do |question, index|
         next unless ask_conditional_question?(question)
-        yield(question, index + 1)
+        progress = "(#{index + 1}/#{questions_length})"
```
bobwhitelock: Can we add back the `progress_indicator` method for defining this to this class? This would be a bit clearer I think.

src/configurator/question.rb
```diff
@@ -59,21 +59,12 @@ def use_readline?
...
-        # Default for a boolean question which has a previous answer should be
-        # set to the input HighLine's `agree` expects, i.e. 'yes' or 'no'.
-        current_answer_value ? 'yes' : 'no'
-      end
-
-      # The answer value this question at this level would currently take.
-      def current_answer_value
-        old_answer.nil? ? default : old_answer
+        return nil if default.nil?
+        return (default ? 'yes' : 'no') if type.boolean?
```
bobwhitelock: I don't think inlining this is an improvement as it makes this method quite dense. Changing this to

```ruby
def default_input
  type.boolean? ? boolean_default_input : default
end
```
and adding back `boolean_default_input` as it was before would better separate what handling here is only needed for boolean questions (special handling for `nil` and using `yes` or `no`) from everything else (just using `default` as the default).

src/validation/loader.rb
```diff
@@ -41,18 +41,14 @@ def question_tree
...
 
-      # Returns a tree
+      # TODO: Remove this method
       def configure_section(section)
-        question_tree.children.find { |c| c.name == section }
+        question_tree.section_tree(section)
       end
 
-      # Returns a hash of identifiers and questions
+      # TODO: Remove this method
```
bobwhitelock: If you think we should remove these 2 methods can you just remove them, or document why you think they should be removed but haven't been yet if this isn't currently possible? Otherwise the codebase just becomes littered with cruft.

src/question_tree.rb
```diff
@@ -27,14 +25,17 @@ class QuestionTree < Tree::TreeNode
...
     end
 
     def questions_length
-      num = 0
-      filtered_each { |_q| num += 1 }
-      num
+      @questions_length ||= begin
+        num = 0
+        filtered_each { |_q| num += 1 }
+        num
```
bobwhitelock: Rather than re-implementing `Enumerable#length` can we just use it, like this:

```ruby
def questions_length
  @questions_length ||= filtered_each.to_a.length
end
```
?

The tests pass when I make this change so I presume this works fine.

spec/hash_mergers/hash_merger_spec.rb
WilliamMcCumstie: Done

src/question_tree.rb
WilliamMcCumstie: Done

src/question_tree.rb
WilliamMcCumstie: Done

src/question_tree.rb
WilliamMcCumstie: I'm not sure, it's an additional method in an already long class. Also the `index` isn't known in advance so it would still need to be an input.

I have however renamed the `progress` variable to `progress_indicator`

src/configurator/question.rb
WilliamMcCumstie: Done

src/question_tree.rb
WilliamMcCumstie: Done

src/validation/loader.rb
WilliamMcCumstie: Removed

PR #340, opened 2018-03-12
Fix saving of the answers

### LINKED ISSUE #330: [2018.1.0-rc7] Configure Defaults vs Domain Level Inheritance
I have the following setup in `configure.yaml` for one of the questions:
```
  - &pri_network_interface
    identifier: pri_network_interface
    question: 'Interface for the primary network'
    default: 'eth0'
```

When I run through `metal configure domain` I set the interface to `em1`, which gives me the following in `answers/domain.yaml`:
```
pri_network_interface: em1
```

When running through `metal configure local` I see the domain-level answer as the one it would use:
```
Interface for the primary network (3/60)  |em1|
eth0
```

But if I change this to `eth0` (the default in `configure.yaml`) I get the following in `answers/nodes/local.yaml` (note how the interface isn't specified):
```
pri_network_defined: true
pri_network_ip: 10.10.0.1
```

This means that it ends up inheriting the domain answer instead of using the configure default, thus, meaning I can't use `eth0` unless it's specified at the domain level to match the default:
```
[root@localhost metalware]# metal each local 'echo <%= answer.pri_network_interface %>'
em1
```

This also effects the groups which I configure to use the domain default, the answers file will specify the interface name again, this could cause a lot of confusion with inheritance (below is `answers/groups/nodes.yaml`):
```
pri_network_defined: true
pri_network_interface: em1
```

WilliamMcCumstie:
Based on #339, please merge it first. This PR fixes #330.

This PR ensures that the configured answer is always picked and returned from the namespace. This is different to whether the answer is saved. Due to the inheritance of answers from one level to another, the configured answer does not get saved if it matches its higher answer. Despite this, the inheritance ensures the configured answer is still returned.

Most of this PR concerns testing edge cases for each for the following levels: `group`, `node`, and `local` as well as ensuring it doesn't fail for orphan nodes.


bobwhitelock:

spec/namespaces/node_spec.rb
```diff
@@ -238,17 +238,25 @@ def local_node_uses_local_build?(config_build_method)
...
+      AlcesUtils.spoof_nodeattr(self)
+
+      # Enable/disable plugins for node as needed.
+      enabled_identifier = \
+        Metalware::Plugins.enabled_question_identifier(enabled_plugin)
+      disabled_identitifer = \
+        Metalware::Plugins.enabled_question_identifier(disabled_plugin)
+      answers = {
+        enabled_identifier => true,
+        disabled_identitifer => false,
```
bobwhitelock: These names are slightly confusing - these are both 'enabled identifiers', the former is just the enabled identifier for an enabled plugin and the latter the enabled identifier for a disabled plugin. `enabled_plugin_identifier` and `disabled_plugin_identifier` would be more accurate.

PR #342, opened 2018-03-13
Do not save orphan until the end of the configure process

### LINKED ISSUE #341: Orphan node added to group cache even if configure is cancelled
Evidence of orphan node being present despite the setup being cancelled
```
[root@controller [testcluster4] ~]# metal configure node randomnode1
warning: Could not find node 'randomnode1' in genders file. The node will be added to the orphan group.

The node will not be removed from the orphan group automatically. The behaviour of an orphan node that is later added to a group is undefined. A node can be removed from the orphan group by editing:
/var/lib/metalware/cache/groups.yaml
Root user encrypted password (1/61)  |$1$e0deogOd$/baLGKZo1ApjqX28aW31X.|

Kickstart template to use (2/61)  |default|

PXELINUX template to use (3/61)  |default|

Setup primary network? (4/61) [yes/no]
y
IP for nodes on the primary network (5/61)  |10.10.<%= node.group.index %>.<%= node.index %>|
Exiting without saving...
[root@controller [testcluster4] ~]# cat /var/lib/metalware/cache/groups.yaml
---
next_index: 3
primary_groups:
  nodes: 1
  infra: 2
orphans:
- local
- randomnode1
```
It seems it gets added to the group cache at the beginning of the command. 

I've tested cancelling a group configure command and it does not add the group.

WilliamMcCumstie:
Based on #340 , please merge it first.

Previously the orphan nodes where being saved to the `group_cache` right at the beginning of the configure process. This meant exiting it early still resulted in the orphan being added.

The fix is not to implicitly save the `GroupCache` after a node is added. Instead a new `GroupCache.update` method has been created, which can be used instead of `new`. It will yield an instance of the cache and save it at then end. This means if the process exits early, the cache is not saved.

Fixes #341

### DIFF
--- a/src/commands/configure/group.rb
+++ b/src/commands/configure/group.rb
@@ -36,7 +36,6 @@ module Metalware
 
         def setup
           @group_name = args.first
-          @cache = GroupCache.new
         end
 
         def configurator
@@ -49,7 +48,7 @@ module Metalware
         end
 
         def custom_configuration
-          cache.add(group_name)
+          GroupCache.update { |c| c.add group_name }
         end
       end
     end
--- a/src/group_cache.rb
+++ b/src/group_cache.rb
@@ -29,6 +29,10 @@ module Metalware
   class GroupCache
     include Enumerable
 
+    def self.update
+      yield new
+    end
+
     def initialize(force_reload_file: false)
       @force_reload = force_reload_file
     end
--- a/src/group_cache.rb
+++ b/src/group_cache.rb
@@ -30,7 +30,9 @@ module Metalware
     include Enumerable
 
     def self.update
-      yield new
+      cache = new
+      yield cache
+      cache.save
     end
 
     def initialize(force_reload_file: false)
@@ -45,13 +47,11 @@ module Metalware
       return if group?(group)
       primary_groups_hash[group.to_sym] = next_available_index
       bump_next_index
-      save
     end
 
     def remove(group)
       pgh = primary_groups_hash
       pgh.delete(group.to_sym)
-      save
     end
 
     def each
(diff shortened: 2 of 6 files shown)


bobwhitelock:
Is this a fix for #341?

WilliamMcCumstie:
Opps yes it is

bobwhitelock:

src/group_cache.rb
```diff
@@ -29,6 +29,12 @@ module Metalware
   class GroupCache
     include Enumerable
 
+    def self.update
```
bobwhitelock: `update` as a name for this could maybe be clearer, as it's not clear that this should take a block and won't update the cache immediately - how about `transaction`, as it functions similarly to a database transaction? I.e. it will either save the cache if everything succeeds or not save it (rollback to the previous state) if any error occurs.

src/commands/configure/group.rb
```diff
@@ -49,7 +48,7 @@ def answer_file
         end
 
         def custom_configuration
-          cache.add(group_name)
+          GroupCache.update { |c| c.add group_name }
```
bobwhitelock: Both here and in the tests we do similar things to this in several places, where we call `GroupCache.update` only to `add` a single node in the block. This seems a bit too verbose for what we're doing, and looks slightly confusing as it's not clear why we need the `update` if we're not doing anything else in the block passed to it.

A possible improvement would be to add an `add` class method to `GroupCache` to abstract away this pattern. This would then make it clearer I think that `GroupCache.add` will just immediately add a node to the group cache, whereas `GroupCache.transaction do ...` (new name as suggested below) will allow adding nodes to the group cache within a transaction which may or may not be saved.

PR #346, opened 2018-03-14
Fix merging of `configure.yaml` defaults

WilliamMcCumstie:
Based on #343, please merge it first. This PR fixes #345.

In an attempt to ensure all questions use the same yaml defaults regardless in level, the defaults from non-domain levels where being ignored. This meant that questions that do not get asked at the domain level can not have a default at all.

Instead the yaml defaults at each level are merged in reverse priority order to form the root defaults for all questions. This ensures each question can have a default BUT which default is used is independent of section level.

Also fixes #347 which in the end being an easy fix (see https://github.com/alces-software/metalware/pull/346/commits/6736efdfcf03a58242af3d4d983cb29b67fae7d4). The error was caused by the rendering of the answers during the save process. However the save process should be using the un-rendered answers anyway. So before the comparison, the existing answers are converted to a regular hash that causes the render to be skipped.

PR #349, opened 2018-03-14
Add ipmi command flags back to the cli

### LINKED ISSUE #348: Change of method for passing IPMI command to be used in `metal ipmi`
Originally the `alces ipmi` and `metal ipmi` tools used `-c` to pass the command to be used in the underlying `ipmitool`. This was (from what I can see..) to emulate the behaviour of `ipmitool` which takes `-c` for the command to be sent to either the local IPMI device or a remote IPMI interface (when using `-H` to `ipmitool`.

This has changed to `-k` and most recently nothing supplied for the command. Not having to remember which version of `metal ipmi` or `alces ipmi` is being used for a particular cluster would be good as I can now think of 5 or 6 different ways of performing the same action throughout the different generations of cluster.

WilliamMcCumstie:
Based on #346, please merge it first.

This fixes #348 which raise a very good point that the command input to `ipmi` has been highly volatile. This has been the case for various reasons, without going into depth. However to keep the current `UI` consistent with passed versions, the `-k`, `-c` and `--command` flags have been added back into the `cli`.

They serve no purpose internally within metalware as the command can be inferred by commander automatically. However including them in the command will no longer raise an error.

It does mean however the following command will still error:
```
metal ipmi -c COMMAND -g GROUP_NAME
```
Because the `-c` flag is ignored, metalware will interpret `COMMAND` as the group input and `GROUP_NAME` as the command input. Whilst this might be unexpected, the first input is consistently used as the node/group input into the other `metal` commands (where applicable).

It would be possible to fix this, however it would require adding additional logic to metalware internally rather than updating the `cli`.

### DIFF
--- a/src/commands/ipmi.rb
+++ b/src/commands/ipmi.rb
@@ -43,9 +43,9 @@ module Metalware
           regular_command = (args.length > 1 ? args[1] : nil)
           two_commands = regular_command && options.command
           raise InvalidInput, TWO_COMMANDS_ERROR if two_commands
-          (regular_command || options.command).tap do |command|
-            raise InvalidInput, NO_COMMAND_ERROR unless command
-          end
+          command = (regular_command || options.command)
+          raise InvalidInput, NO_COMMAND_ERROR unless command
+          command
         end
       end
 
--- a/src/commands/ipmi.rb
+++ b/src/commands/ipmi.rb
@@ -31,13 +31,13 @@ module Metalware
   module Commands
     class Ipmi < CommandHelpers::BaseCommand
       class Command
-        TWO_COMMANDS_ERROR = <<~EOF.squish
+        TWO_COMMANDS_ERROR = <<~ERROR.squish
           The COMMAND input can not be used with a LEGACY_COMMAND flag
-        EOF
+        ERROR
 
-        NO_COMMAND_ERROR = <<~EOF.squish
+        NO_COMMAND_ERROR = <<~ERROR.squish
           No command given. Use --help for more information
-        EOF
+        ERROR
 
         def self.parse(args, options)
           regular_command = (args.length > 1 ? args[1] : nil)
--- a/src/commands/ipmi.rb
+++ b/src/commands/ipmi.rb
@@ -71,7 +71,13 @@ module Metalware
 
       private
 
+      attr_reader :command_argument
+
       prepend CommandHelpers::NodeIdentifier
+
+      def setup
+        @command_argument = Command.parse(args, options)
+      end
 
       def run
         nodes.each do |node|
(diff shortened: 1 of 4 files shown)


bobwhitelock:

bobwhitelock:
I think adding back `-c`/`-k`/`--command` as ways to pass the `ipmi` command to run is a good idea, to make things simpler to run for users. However, I think this issue with this particular way of adding these back is likely to cause more confusion than adding them back solves:

> It does mean however the following command will still error:
> ```
> metal ipmi -c COMMAND -g GROUP_NAME
> ```
> Because the `-c` flag is ignored, metalware will interpret `COMMAND` as the group input and `GROUP_NAME` as the command input. Whilst this might be unexpected, the first input is consistently used as the node/group input into the other `metal` commands (where applicable).

as this means that a valid command in one version of Metalware will give very odd behaviour when run on a later version of Metalware.

A better approach I think would be to:

- make `COMMAND` an optional argument;

- handle the various possibility option aliases for the `COMMAND` argument (`-c`/`-k`/`--command`) in such a way that each is explicitly handled in the same way as the argument, rather than them just having these happen to be handled correctly if the arguments are provided in the correct order.

We already have a method where we unpack the command to use, so we can add additional handling for these other ways of providing the command to this:

https://github.com/alces-software/metalware/blob/b8b83afe5897db1ef804241862c64fb697d70cb2/src/commands/ipmi.rb#L71-L73

For added robustness we could also show a user input error if no method of providing the command is given, or if multiple are.

WilliamMcCumstie:
Okay, I have fully integrated the legacy command inputs into metalware. Now `COMMAND` xor `LEGACY_COMMAND` can be used with `ipmi`. An error is raised if both or neither are included.

src/commands/ipmi.rb
```diff
@@ -69,7 +88,7 @@ def ipmi_command(node, arguments:)
       end
 
       def command_argument
-        args[1]
+        Command.parse(args, options)
```
bobwhitelock: Because we're not attempting to parse the command argument/option until we use `command_argument`, which happens after other parts of the `ipmi` command such as loading parts of the node's config, this can lead to additional possibly confusing output:

```
[root@controller(dev) metalware]# bin/metal ipmi node01
warning: The following returned nil: answer.is_vm
warning: The following returned nil: answer.libvirt_host
error: No command given. Use --help for more information.
```

Possibly we should parse the `command_argument` in a `setup` method, so that the command argument is either successfully parsed up-front or the command fails, rather than having additional warnings potentially printed before informing the user that their command line itself is invalid. `command_argument` could then be accessed via an `attr_reader`.

bobwhitelock:
Made the needed changes myself (as you're away) so we can put out a release sooner rather than later. Thanks for looking into this, will merge once CI passes (+/- some flaky tests).

PR #356, opened 2018-04-03
Create planning document for asset records

WilliamMcCumstie:
Feel free to make suggestions

### DIFF
--- a/docs/design/04-asset-records.md
+++ b/docs/design/04-asset-records.md
@@ -29,9 +29,9 @@ The template will then be opened using the system editor (see issues). Before
 being saved to the final location in `/var/lib/metalware/assets`. See
 `Editing` for more details on this process.
 
-In short, the validation will occur when the file is saved. This means the
-initial "templates" do not need to be valid `YAML`. Be careful of the file
-extension as it might not be `.yaml` (see issues).
+In short, the validation will occur when the file is saved. Currently the
+template file must be valid `YAML`. Blank fields will therefore need to be
+represented with `nill`.
 
 As this is a new record, the `type` is required for the command. This tells
 `metalware` which "template" should be used. The templates should be stored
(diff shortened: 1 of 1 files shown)


docs/design/04-asset-records.md
```diff
@@ -0,0 +1,131 @@
...
+At this stage re-editing a previously bad file does not need to be supported.
+This can be added in the future.
+
+At the end, the temporary file is moved to its final location.
+
+# Accessing Asset Records and Naming
+
+The asset records will be loaded into the namespace under `alces.assets`.
+They will be accessible by name as both: `alces.assets.<name>` and
+`alces.assets.find_by_name(name)`. The `MetalArray` could be used for this
```
bobwhitelock: Is there a benefit of having the `alces.assets.find_by_name('name')` form as opposed to running `alces.assets.name` directly?

docs/design/04-asset-records.md
```diff
@@ -0,0 +1,131 @@
...
+that will provide the basis for the asset record. An no point will the
+entire "template" be rendered using ERB (however the individual tags might?).
+
+The template will then be opened using the system editor (see issues). Before
+being saved to the final location in `/var/lib/metalware/assets`. See
+`Editing` for more details on this process.
+
+In short, the validation will occur when the file is saved. This means the
+initial "templates" do not need to be valid `YAML`. Be careful of the file
+extension as it might not be `.yaml` (see issues).
```
bobwhitelock: Is it necessary that the templates should be able to not initially be valid YAML? If we require these to be valid YAML initially, possibly with `null` entries for fields which have no other valid default, then:

- it will be easier for users initially creating/editing the template to tell they have done this correctly, as valid YAML => it's a valid template (for now);

- related to this, we can validate the template on load as well as save, which will avoid the potential of users being confused after saving a template and this failing, if the template itself contained an error they didn't spot on load;

- it will be slightly easier to initially edit a template to create a valid asset YAML file, as you can find all `null` entries in your `$EDITOR` and replace them with new values.

bobwhitelock:
Also your docs have failing tests :stuck_out_tongue:.

WilliamMcCumstie:
Rebased on develop to hopefully fix the failing tests.

docs/design/04-asset-records.md
WilliamMcCumstie: The two methods for accessing have slightly different purposes. If you know the name in advanced and it can be hard coded, then `alces.assets.<name>` is preferred.

However if the name is a variable or won't be known at runtime then `send` would have to be used to get the asset. As this could have side affects, a more robust `find_by_name` method is preferred.

So having both methods is ideal depending on the circumstances. This matches the accessing methods used for `alces.groups` and `alces.nodes`.

docs/design/04-asset-records.md
WilliamMcCumstie: Either method is easy enough to implement. It more comes down to what @ste78 and @ColonelPanicks prefers.

WilliamMcCumstie:
Thanks for the review. As far as the associations between the records go, the requirement for it is still kinda vague. It is not an essential feature for the initial implementation, so we can kick the issue down the road for a bit.

docs/design/04-asset-records.md
ColonelPanicks: I'm happy to include 'null' as default values in my templates. Will have some templates in metalware-default today.

ste78:
all looks bang on to me. also i’m not sure how we’re gonna do the relationships as well but I totally blame whoever is coming up with the requirements 😬

Be good to get the basics in so we can see where we go from there.

Issue #275, opened 2018-01-03
Default values aren't displayed when the `type` is `Integer`


vlj91:
As above, if the `type` of a question is `Integer`, the default value isn't displayed when running `configure`:

```
Enter the IP or hostname of the Libvirt host (20/25)  |julius|

Amount of memory to allocate to a VM in GiB (21/25)

Number of CPU cores to allocate to a VM (22/25)

Size of the VM root disk to allocate in GB (23/25)

Specify the Libvirt disk pool name (24/25)  |vm|
```

```yaml
  - &vm_memory
    identifier: vm_memory
    question: 'Amount of memory to allocate to a VM in GiB'
    type: 'integer'
    default: 4

  - &vm_cpus
    identifier: vm_cpus
    question: 'Number of CPU cores to allocate to a VM'
    type: 'integer'
    default: 1
```

PR #357, opened 2018-04-04
Correctly display integer values during the configure process

atoghill:
This pr implements a fix for #275 by simply converting the default to a string


WilliamMcCumstie:
LGTM. If there are issues with defaults not being displayed proper in the future, we should look into the monkey patching of `HighLine`. However a small fix is suitable for now.

This can be merged once the `CI` finishes running

PR #358, opened 2018-04-04
Create an utility for opening files in an editor

WilliamMcCumstie:
Opening files in an editor will become a key component of #356.

This utility takes a file path and opens it in the system editor according to `$VISUAL` or `$EDITOR`. If neither are set, then `vi` is used instead.

Most of the commits are concerning the tests. Note that the editor command is not ran during the spec to prevent it opening. Opening the editor during the spec does not play well with `rspec` and creates issues with swap files.

Instead the command is expected to be called with the correct arguments. The ability to open and close the editor has been tested manually. As this action is blocking, as long as the correct command is called the outcome shouldn't change.

### DIFF
--- a/spec/utils/editor_spec.rb
+++ b/spec/utils/editor_spec.rb
@@ -4,6 +4,7 @@ require 'utils/editor'
 
 RSpec.describe Metalware::Utils::Editor do
   subject { Metalware::Utils::Editor }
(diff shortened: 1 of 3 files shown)


WilliamMcCumstie:
Rebased on develop to fix failing tests.

PR #360, opened 2018-04-05
Adding paths to asset files

atoghill:
* Implements the two methods that define the paths for asset templates and save locations; `asset_template` and `asset_final` respectively.
* Adds a `spec` for testing these methods return correct paths
* Replaces any hard coded paths within `FilePath` that simply replicate other method calls with the appropriate method
* Edits the `install` script and `FileSystem` class to include the `/v/l/m/assets` dir in their processes

### DIFF
--- a/spec/asset_path_spec.rb
+++ b/spec/asset_path_spec.rb
@@ -1,0 +1,15 @@
+require 'file_path'
+
+RSpec.describe Metalware::FilePath do
+  describe 'asset files' do
+    let :file_path { Metalware::FilePath }
+
+    it 'defines an asset template' do
+      expect(file_path.asset_template('rack')).to eq('/var/lib/metalware/repo/assets/rack.yaml')
+    end
+
+    it 'defines an asset final save path' do
+      expect(file_path.asset_final('rack01')).to eq('/var/lib/metalware/assets/rack01.yaml')
+    end
+  end
+end
--- a/spec/asset_path_spec.rb
+++ b/spec/asset_path_spec.rb
@@ -5,11 +5,13 @@ RSpec.describe Metalware::FilePath do
     let :file_path { Metalware::FilePath }
 
     it 'defines an asset template' do
-      expect(file_path.asset_template('rack')).to eq('/var/lib/metalware/repo/assets/rack.yaml')
+      path = File.join(file_path.repo, 'assets', 'rack.yaml')
+      expect(file_path.asset_template('rack')).to eq(path)
     end
 
     it 'defines an asset final save path' do
-      expect(file_path.asset_final('rack01')).to eq('/var/lib/metalware/assets/rack01.yaml')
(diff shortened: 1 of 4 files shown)


spec/asset_path_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'file_path'
+
+RSpec.describe Metalware::FilePath do
+  describe 'asset files' do
+    let :file_path { Metalware::FilePath }
+
+    it 'defines an asset template' do
+      expect(file_path.asset_template('rack')).to eq('/var/lib/metalware/repo/assets/rack.yaml')
```
WilliamMcCumstie: These lines are quite long, they should stick to max 80 characters. Something like:
```
path = /v/l/m/...
expect(file_path.asset_template('rack').to eq(path)
```

src/file_path.rb
```diff
@@ -148,6 +148,14 @@ def log
         '/var/log/metalware'
       end
 
+      def asset_template(type)
+        File.join('/var/lib/metalware/repo/assets', "#{type}.yaml")
```
WilliamMcCumstie: In this case, it probably would be clearer to go `type + '.yaml'`

WilliamMcCumstie:
Also `enhancement` is a non-standard branch name. There is nothing inherently wrong with it, except it is long to type.

Typically for something new, you would make a `feature` branch. If you are developing something but not adding a new feature, `dev` is also fine. Bug fixes are denoted by `fix` or if it is in a release and needs to be fixed urgently then use `hotfix`. Finally `wip` (work in progress) or `junk` can be use if you just want a test branch (`dev` also works here).

src/file_path.rb
```diff
@@ -148,6 +148,14 @@ def log
...
+      end
+
+      def asset_final(name)
```
WilliamMcCumstie: Also can you change the name of this method to just `asset`. Whilst this is the final location for the saving process, they will be loaded again latter.

Calling `FilePath.asset` is descriptive enough for our purposes.

src/file_path.rb
WilliamMcCumstie: Also the `/v/l/m/repo` shouldn't be hard coded. You can refer to it using the `repo` method.

atoghill:
@WilliamMcCumstie Made changes as requested

src/file_path.rb
```diff
@@ -117,7 +117,7 @@ def rendered_build_file_path(rendered_dir, section, file_name)
       end
 
       def rendered_files
-        '/var/lib/metalware/rendered'
+        Constants::RENDERED_DIR_PATH
```
WilliamMcCumstie: As you have defined this method as a constant, it can be removed. `RENDERED_DIR_PATH` will be dynamically loaded in as `rendered_dir`

atoghill: Yep I missed that! Fixed

PR #361, opened 2018-04-09
Edit/Copy Utility

atoghill:
This PR adds the tests and functionality for the `open_copy` method with the purpose of copying and/or editing files


atoghill:
Note: There is a slight overlap with the two tests. I could roll them into one singular tests which would eliminate the repetition. However I felt it was important to test these two aspects separately, if there is a better/easier way to do this then I'm all ears

PR #363, opened 2018-04-10
Asset Cache

atoghill:
This PR implements functionality and tests for assigning assets to nodes using an asset cache file. It currently stores the relationship in the following format:
```
node:
  node001: asset001
  node002: asset004
  ...
```
## Methods
`data`
* Loads the data and caches it internally
* Initially the file is empty so it creates the necessary structure

`save`
* Saves the data back to `YAML`

`assign_asset_to_node(asset_name, node)`
* Taking a `String` and `Namespaces::Node` as inputs respectively
* Updates the `node` key within the cache file `assets.yaml` using the `asset_name` given as it's value

`asset_for_node(node)`
* Takes `Namespaces::Node` as input
* Returns the value of the key given by `node`

### DIFF
--- a/spec/asset_cache_spec.rb
+++ b/spec/asset_cache_spec.rb
@@ -1,0 +1,53 @@
+require 'alces_utils'
+require 'cache/asset'
+
+RSpec.describe Metalware::Cache::Asset do
+  include AlcesUtils
+  
+  let :cache { Metalware::Cache::Asset.new }
+  let :content { { node: { node_name.to_sym => 'asset_test' } } }
+  let :node_name { 'test_node' } 
+  let :node { alces.nodes.find_by_name(node_name) }
+
+  AlcesUtils.mock(self, :each) do
+    mock_node(node_name) 
+  end
+  
+  describe '#data' do
+    it 'handles empty cache' do
+      expect do
+        cache.data
+      end.not_to raise_error
+    end
+
+    it 'returns populated cache' do
+      Metalware::Data.dump(Metalware::FilePath.asset_cache, content)
+      expect(cache.data).to eq(content)
+    end
+  end
+
+  describe '#save' do
+    it 'saves the cache to yaml' do
+      cache.assign_asset_to_node('asset_test', node)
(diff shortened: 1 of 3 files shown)


spec/asset_cache_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      expect do
+        cache.assign_asset_to_node('asset_test', node)
+      end.not_to raise_error
+    end 
+  end
+
+  describe '#asset_for_node' do
+    it 'returns the corresponding nodes asset' do
+      cache.assign_asset_to_node('asset_test', node)
+      cache.save
```
WilliamMcCumstie: The save is superfluous in this test. `asset_for_node` should return the asset regardless if it has been saved

src/cache/asset.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'file_path'
+require 'data'
+
+module Metalware
+  module Cache
+    class Asset
+      def data
+        @assets ||= begin
```
WilliamMcCumstie: As the is the cache for the `data` method, the instance variable should also be called data (as opposed to `asset`)

src/cache/asset.rb
```diff
@@ -0,0 +1,36 @@
...
+          Data.load(FilePath.asset_cache).tap do |a|
+            a.merge! blank_cache if a.empty?
```
WilliamMcCumstie: Try not using `merge!` if it isn't required, it can have some nasty side affects if you aren't careful. I am aware a lot of existing code is implemented in this manor, however this should also be changed at some point (ideally).

In this case, something along these lines will work:
```
raw_load = Data.load(path)
raw_load.empty? ? blank_cache : raw_load
```

atoghill:
@WilliamMcCumstie made the requested changes and there are no complaints from our favourite cop and/or Travis. With your approval I'd be happy to merge this :)

WilliamMcCumstie:
Thanks for those changes

PR #364, opened 2018-04-10
Implement the add asset command

WilliamMcCumstie:
The `metal asset add` command takes template files from the repo and allows the user to edit them before saving. It is mostly a wrapper around `Editor.open_copy` with a few file checks to ensure the template exists and the asset doesn't.

The `metal asset edit` command has also been added into the `cli` and the file has been create. It otherwise has not been implemented.

The `render_erb_template` method in the `HashMergerNamespace` has also been simplified. This is otherwise unrelated but needed to be done at some point.

### DIFF
--- a/spec/commands/asset/add_spec.rb
+++ b/spec/commands/asset/add_spec.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'commands'
+require 'utils'
+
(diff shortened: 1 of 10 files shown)


atoghill:
Looks good to me, once rubocop is happy then I think it should be fine to merge

PR #367, opened 2018-04-12
Implement the asset delete command

atoghill:
`asset delete` allows the user to unassign an asset from a node and then delete it. If the asset trying to be deleted does not exist then the user is presented with an error message to inform them of this and no changes will be made to the `assets/yaml` cache file.


WilliamMcCumstie:
This PR has a few bugs that have been fixed in #368. Thus the two PRs will be reviewed as one in the next commit.

PR #368, opened 2018-04-12
Access asset records from the node namespace

atoghill:
This PR implements methods to the `alces_static` and `node` namespaces to allow for the loading of the asset cache and node's asset respectively.


atoghill:
All changes within this branch, including the requested changes in the review, are within the most recent branch found within #370. Therefore it is best to review the two together there

PR #370, opened 2018-04-13
Show the link to the asset when calling 'alces.node.asset'

atoghill:
When calling `alces.node.asset` an object should now be returned containing two keys, `name` and `data`. This branch is based off of #368

### DIFF
--- a/spec/cache/asset_spec.rb
+++ b/spec/cache/asset_spec.rb
@@ -59,4 +59,12 @@ RSpec.describe Metalware::Cache::Asset do
       expect(new_cache.asset_for_node(node)).to eq(asset_name)
     end
   end
+
+  describe '#unassign_asset' do
+    it 'unassigns the asset from the node' do
+      cache.unassign_asset('asset_test')
+      new_cache = Metalware::Cache::Asset.new
+      expect(new_cache.data).not_to eq(content)
+    end
+  end
 end
--- a/spec/cache/asset_spec.rb
+++ b/spec/cache/asset_spec.rb
@@ -66,5 +66,11 @@ RSpec.describe Metalware::Cache::Asset do
       new_cache = Metalware::Cache::Asset.new
       expect(new_cache.data).not_to eq(content)
     end
+
+    it 'attempts to unassign a missing asset' do
+      cache.unassign_asset('missing_asset')
+      new_cache = Metalware::Cache::Asset.new
+      expect(new_cache.data).not_to eq(content) 
+    end
   end
 end
--- a/spec/commands/asset/delete_spec.rb
+++ b/spec/commands/asset/delete_spec.rb
@@ -22,9 +22,7 @@ RSpec.describe Metalware::Commands::Asset::Delete do
 
   context 'when using a saved asset' do
     before do
-      FileSystem.root_setup do |fs|
-        fs.with_minimal_repo
-      end
+      FileSystem.root_setup(&:with_minimal_repo)
     end
 
     let :asset_path { Metalware::FilePath.asset(asset) }
--- a/spec/commands/asset/delete_spec.rb
+++ b/spec/commands/asset/delete_spec.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+require 'cache/asset'
+require 'filesystem'
+require 'commands'
+require 'utils'
+
+RSpec.describe Metalware::Commands::Asset::Delete do
+
+  it 'errors if the asset doesnt exist' do
+    expect do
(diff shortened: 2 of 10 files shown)


spec/cache/asset_spec.rb
```diff
@@ -59,4 +59,12 @@
       expect(new_cache.asset_for_node(node)).to eq(asset_name)
     end
   end
+
+  describe '#unassign_asset' do
+    it 'unassigns the asset from the node' do
+      cache.unassign_asset('asset_test')
+      new_cache = Metalware::Cache::Asset.new
+      expect(new_cache.data).not_to eq(content)
+    end
```
WilliamMcCumstie: Can we also add a spec for what happens if the asset is missing? Mainly ensure that it doesn't error

spec/commands/asset/delete_spec.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require 'cache/asset'
+require 'filesystem'
+require 'commands'
+require 'utils'
+
+RSpec.describe Metalware::Commands::Asset::Delete do
+
```
WilliamMcCumstie: Remove this empty line

spec/commands/asset/delete_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  it 'errors if the asset doesnt exist' do
```
WilliamMcCumstie: So the `doesnt` needs the apostrophe. Either convert the string quotes to double quotes, or better yet, expand the contraction to `does not`.

spec/commands/asset/delete_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    expect do
+      Metalware::Utils.run_command(described_class,
+                                   'missing-type',
+                                   'name',
```
WilliamMcCumstie: The `delete` command should only accept a single input. However this spec contains two: `'missing-type', 'name'`. Thus it isn't clear why this spec if failing.

Also a bit further down a `run_command` method has been defined. This spec is currently duplicating that code. This should be condensed down.

spec/commands/asset/delete_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+                                   stderr: StringIO.new)
+    end.to raise_error(Metalware::InvalidInput)
+  end
+
+  context 'when using a saved asset' do
+    before do
+      FileSystem.root_setup do |fs|
```
WilliamMcCumstie: This can be simplified to: `FileSystem.root_setup(&:with_minimal_repo)`.

src/command_helpers/asset_helper.rb
```diff
@@ -46,6 +46,13 @@ def raise_error_if_node_is_missing
           Can not find node: #{options.node}
         EOF
       end
+
+      def error_if_asset_file_doesnt_exist(asset_name, asset_path)
+        return if File.exist?(asset_path)
+        raise InvalidInput, <<-EOF.squish
+          The "#{asset_name}" asset does not yet exist
```
WilliamMcCumstie: Remove the `yet` from this comment. It makes it sound like, the asset should be added, which doesn't make sense if we are trying to delete it.

src/command_helpers/asset_helper.rb
```diff
@@ -46,6 +46,13 @@ def raise_error_if_node_is_missing
           Can not find node: #{options.node}
         EOF
       end
+
+      def error_if_asset_file_doesnt_exist(asset_name, asset_path)
```
WilliamMcCumstie: Both the `asset_name` and `asset_path` are not require as inputs to this method. The `asset_path` can be found from the name alone.

src/namespaces/node.rb
```diff
@@ -86,6 +86,14 @@ def plugins
         @plugins ||= MetalArray.new(enabled_plugin_namespaces)
       end
 
+      def asset
+        @asset ||= begin
+          asset_name = alces.asset_cache.asset_for_node(self)
+          return unless asset_name
+          OpenStruct.new(name: asset_name, data: alces.asset.find_by_name(asset_name))
```
WilliamMcCumstie: Break this struct definition into multiple lines. It would read clearer as:
```
OpenStruct.new(
  name: ....
  data: .....
)
```

WilliamMcCumstie:
Rebased the merge commit out (https://github.com/alces-software/metalware/commit/a6ff247de646d0569741318590a377a965fdf97a)

PR #373, opened 2018-04-16
Add asset link and unlink commands

atoghill:
This PR implements two new commands for assets, `asset link` and `asset unlink`.
`metal asset link` takes two arguments, `ASSET_NAME` and `NODE_NAME` . This will create a link between the asset and node within `cache/assets.yaml`
`metal asset unlink` takes the same arguments but instead removes the link between the asset and node within the cache file.


spec/cache/asset_spec.rb
```diff
@@ -61,16 +61,30 @@
...
+    before :each do
+      cache.assign_asset_to_node(asset_name, node)
+      cache.save
+    end
+
+    it 'unassigns the asset from all found instances ' do
+      cache.unassign_asset(asset_name)
+      cache.save
+      new_cache = Metalware::Cache::Asset.new
+      expect(new_cache.data).not_to eq(content)
```
WilliamMcCumstie: Using `not_to eq` in this context isn't the best way to test `unassign_asset`. All it tests is that the cache has changed from what ever is contained in `content`.

Side note, `content` should be renamed `initial_content` as the "content" is changing and thus `content` could be multiple different "contents".

These spec do not test if `unassign_asset` is working correctly, as it could have done any number of alterations to the cache and the spec will still pass.

The purpose of the spec is to document the expected behaviour. Currently from the spec only, I can not tell what it is suppose to do.

Possible something along these lines:
```
expected_content = initial_content.dup.tap do |c|
  c.delete(node_name)
end
expect(new_cache.data).to eq(expected_content)
```

Designing the spec likes this will require you to know which node will be delete from the cache in advance. However this is kinda the point of the spec to show how the `unassign` will work.

A similar comment applies for the remaining specs. Consider how similar they are and try and extract the setup as much as possible.

spec/commands/asset/link_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def run_command
+    Metalware::Utils.run_command(described_class,
+                                 asset_name,
+                                 node_name,
+                                 stderr: StringIO.new)
+  end
+
+  it 'errors when the asset does not exist' do
```
WilliamMcCumstie: We have a few different commands that are meant to error if the asset does not exist.

This should be extracted to a `shared_example` which will clean up the specs.

src/cache/asset.rb
```diff
@@ -25,8 +25,16 @@ def asset_for_node(node)
         data[:node][node.name.to_sym]
       end
 
-      def unassign_asset(asset_name)
-        data.delete_if { |_key, value| value == asset_name }
+      def unassign_asset(asset_name, *node_name)
```
WilliamMcCumstie: What is the purpose of accepting an array of `node_names`? Firstly, as it is an array, it should be plural. Secondly, the condition below only every checks the first element of the array.

AFAICS, only the first node will be unassigned and the rest will be ignored. Either accept and array and unassign all the nodes in it OR remove the array input altogether.

If all you want is an optional input, you can use; `node_name = nil`.

src/cache/asset.rb
```diff
@@ -25,8 +25,16 @@ def asset_for_node(node)
...
+        data.each_value do |value|
+          value.delete_if do |node, asset|
+            if node_name.empty?
+              asset == asset_name
+            else
+              node == node_name[0].to_sym && asset == asset_name
```
WilliamMcCumstie: This line is very dense, and also duplicates an equality check. To fix this, the equality checks should be extracted.

As we are only interested in assets of a particular name, the block can start with a guard statement:
` next unless asset == asset_name`

Assuming my above comment has been addressed, than the return value can be obtained with:
`node_name ? (node == node_name) : true`

src/command_helpers/asset_helper.rb
```diff
@@ -26,6 +26,11 @@ def assign_asset_to_node_if_given(asset_name)
         asset_cache.save
       end
 
+      def unassign_asset_from_node_if_given(asset_name, *node_name)
```
WilliamMcCumstie: Same comment as above. The shouldn't be an array unless it will `unassign` multiple nodes from a particular asset.

src/cache/asset.rb
atoghill: Yep only wanted an optional input

PR #388, opened 2018-04-24
Method for updating the asset cache

atoghill:
We have increasingly seen the need to have a way of updating the asset cache in place. This PR adds the `Cache::Asset.update` method to do this.

I have held off from updating the asset commands to utilise this method until the `CommandHelpers::AssetHelper` is cleaned up.


WilliamMcCumstie:
LGTM

PR #389, opened 2018-04-24
Replace AssetHelper with RecordEditor

atoghill:
The main purpose of this PR is to move away from using the mixin `CommandHelpers::AssetHelper` to a more generalised `CommandHelpers::RecordEditor` that can be inherited by any record related command; `asset` and `layout`.

Currently only the `asset` commands exist but `RecordEditor` aims to support `layout` commands when they are eventually thrown into the mix.

As well as adding this new class there has been some spring cleaning to ensure the class only contains methods that are necessary. Therefore old methods such as `unassign_asset_from_cache` that were only used by a single command have been moved into their respective commands.


src/command_helpers/asset_cache.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Metalware
+  module CommandHelpers
+    module AssetCache
+      private
```
WilliamMcCumstie: So the mixin doesn't need the `private`. If the user wishes to make the method provide, they can include it as such in the calling class.

src/command_helpers/ensure_asset_exists.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Metalware
+  module CommandHelpers
+    module EnsureAssetExists
+      def post_setup
+        super
+        error_if_asset_file_does_not_exist(asset_path)
```
WilliamMcCumstie: There is no need to split this into two methods. It would be simple enough to contain it as one.

WilliamMcCumstie: Also this method is passing the `asset_path` however the method actually accepts the `asset_name`. However this should be resolved by combing them into a single method.

src/commands/asset/delete.rb
```diff
@@ -6,24 +6,29 @@
 module Metalware
   module Commands
     module Asset
-      class Delete < Metalware::CommandHelpers::BaseCommand
-        include CommandHelpers::AssetHelper
+      class Delete < CommandHelpers::BaseCommand
+        include CommandHelpers::EnsureAssetExists
```
WilliamMcCumstie: Move the `include` under the private section. This way it is clearer that all these method will be private.

src/commands/asset/delete.rb
```diff
@@ -6,24 +6,29 @@
...
 
         private
 
-        attr_reader :asset_name, :asset_path
+        attr_reader :asset_name, :asset_path, :cache
 
         def setup
           @asset_name = args[0]
           @asset_path = FilePath.asset(asset_name)
+          @cache = Cache::Asset.new
```
WilliamMcCumstie: Do we need to create a `cache` variable when we have the `AssetCache` mixin?

src/commands/asset/edit.rb
```diff
@@ -3,8 +3,8 @@
 module Metalware
   module Commands
     module Asset
-      class Edit < Metalware::CommandHelpers::BaseCommand
-        include CommandHelpers::AssetHelper
+      class Edit < CommandHelpers::RecordEditor
+        include CommandHelpers::EnsureAssetExists
```
WilliamMcCumstie: As above, make this private

src/commands/asset/link.rb
```diff
@@ -3,12 +3,13 @@
 module Metalware
   module Commands
     module Asset
-      class Link < Metalware::CommandHelpers::BaseCommand
-        include CommandHelpers::AssetHelper
+      class Link < CommandHelpers::BaseCommand
+        include CommandHelpers::EnsureAssetExists
+        include CommandHelpers::AssetCache
```
WilliamMcCumstie: As above make these `includes` private

src/commands/asset/link.rb
```diff
@@ -17,8 +18,8 @@ def setup
         end
 
         def run
-          error_if_asset_file_doesnt_exist(asset_path)
-          assign_asset_to_node_if_given(asset_name)
+          asset_cache.assign_asset_to_node(asset_name, node)
```
WilliamMcCumstie: This can use the `asset_cache.update` method

src/commands/asset/unlink.rb
```diff
@@ -3,19 +3,23 @@
...
   module Commands
     module Asset
-      class Unlink < Metalware::CommandHelpers::BaseCommand
-        include CommandHelpers::AssetHelper
-
+      class Unlink < CommandHelpers::BaseCommand
         private
 
-        attr_reader :node_name
+        attr_reader :node_name, :cache
```
WilliamMcCumstie: Can we use the `AssetCache` mixin here?

src/commands/asset/unlink.rb
```diff
@@ -3,19 +3,23 @@
...
+          @cache = Cache::Asset.new
         end
 
         def run
-          unassign_node_from_cache(node_name)
+          unassign_node_from_cache
+        end
+
+        def unassign_node_from_cache
+          cache.unassign_node(node_name)
```
WilliamMcCumstie: This can use the `cache.update` method

atoghill: This was not based on that branch but once this PR is merged the next thing on the list is to update, no pun intended, everywhere that could use the `cache.update` method

src/commands/asset/unlink.rb
atoghill: Yes we can

WilliamMcCumstie:
I've made a few additional changes that combined the helper modules into a single concern and used aliases within the `add` and `edit` command.

PR #391, opened 2018-04-25
Utilise the update cache method

atoghill:
Within this is a small collection of changes to utilise the `Cache::Asset.update` method added in #388. These changes now mean that the cache is saved in place where appropriate. 

With these changes it also became apparent that `asset_cache` was serving no purpose any more and so I have removed it from `HasAssetConcern`

### DIFF
--- a/src/commands/asset/edit.rb
+++ b/src/commands/asset/edit.rb
@@ -6,8 +6,6 @@ module Metalware
       class Edit < CommandHelpers::RecordEditor
         private
 
-        include CommandHelpers::HasAssetConcern
-
         attr_reader :asset_name, :asset_path
 
         alias source asset_path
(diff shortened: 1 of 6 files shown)


src/commands/asset/edit.rb
```diff
@@ -6,6 +6,8 @@ module Asset
       class Edit < CommandHelpers::RecordEditor
         private
 
+        include CommandHelpers::HasAssetConcern
```
WilliamMcCumstie: I don't think the `Edit` command needs to include `HasAssetConcern` explicitly. It should already be included by `RecordEditor`

WilliamMcCumstie:
Rebased this branch on #392 so the `HasAssetConcern` can be removed. Please merge the other branch first.

WilliamMcCumstie:
Rebased on develop

PR #397, opened 2018-04-26
Add asset type to the metadata

atoghill:
The type associated with an asset is now stored in it's metadata, based on the directory the asset is located within.

### DIFF
--- a/src/namespaces/asset_array.rb
+++ b/src/namespaces/asset_array.rb
@@ -16,7 +16,7 @@ module Metalware
         end
 
         def type
-          @type ||= File.basename(File.dirname(path).singularize)
+          @type ||= File.basename(File.dirname(path)).singularize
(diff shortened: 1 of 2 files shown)


src/namespaces/asset_array.rb
```diff
@@ -15,6 +15,10 @@ def name
           @name ||= File.basename(path, '.yaml')
         end
 
+        def type
+          @type ||= File.basename(File.dirname(path).singularize)
```
WilliamMcCumstie: This looks to be fine however it looks a bit odd to `singularize` a file path. I can't say if that will work in all situations. A clearer approach would be:
```
File.basename(File.dirname(path)).singularize
```

WilliamMcCumstie:
This can be merged once Travis has ran

PR #400, opened 2018-04-27
Implement the layout add command

atoghill:
Using `metal layout add LAYOUT_TYPE LAYOUT_NAME` you can create a new layout from the base of a given `type`. These layouts are stored under `/v/l/m/layouts/<type>/<layout-name>`. 

This command uses the same structure as `asset add` and utilises the same sort of validation thanks to the `Records` class. Therefore you're unable to add from a non-existent type or use a layout name that would cause a conflict, from an existing layout or from any reserved names.

Also within this branch is a new shared example for testing, `record_add_command`, that groups together all tests that are shared between both `asset add` and `layout add`. Because of this changes have been made to the test spec for `asset add` to make use of this new shared example.

### DIFF
--- a/spec/commands/asset/add_spec.rb
+++ b/spec/commands/asset/add_spec.rb
@@ -1,64 +1,14 @@
 # frozen_string_literal: true
 
-require 'commands'
-require 'utils'
-require 'filesystem'
 require 'shared_examples/asset_command_that_assigns_a_node'
+require 'shared_examples/record_add_command'
 
 RSpec.describe Metalware::Commands::Asset::Add do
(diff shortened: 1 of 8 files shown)


atoghill:
The similar code found by `codeclimate` here was expected and can be ignored. As things are going to be heavily changed in the next phase. It seemed unnecessary to group these two similar methods together somewhere temporarily and if necessary can be optimised after the next phase is finalised.

atoghill:
@WilliamMcCumstie See latest commits for the requested changes

WilliamMcCumstie:
LGTM

PR #402, opened 2018-04-27
Add a utility for finding an asset type or layout by name

atoghill:
The `base_path` method within `Records::Layout` checks the given input within the list of `types` and existing `layouts`. Returning the respective path if something is found or raising an error if nothing is found in either place.

From there the path can be used as the base of the asset to be created. This utility was made in preparation for the upcoming changes to `asset add` that allow an asset to be created from a layout or a type.


WilliamMcCumstie:
LGTM, It can be merged once the `CI` has ran

PR #403, opened 2018-04-30
Ensure the source type is valid before making a directory

atoghill:
This implements a fix that ensures the input `ASSET_TYPE` for `metal layout add` is valid before the directory is made. Currently the validity check happens after the directory for the type is made, before it even knows whether or not the type exists.


WilliamMcCumstie:
LGTM

PR #404, opened 2018-04-30
Create `AssetBuilder` that will be used to create assets

WilliamMcCumstie:
A new `AssetBuilder` class has been added to manege the creation of assets and sub assets. The key feature of the class is an internal `queue` that will store the assets that need to be created.

By using this `queue`, each asset can be created in order as well as scheduling new assets to be made.

A new asset can be scheduled using the `push_asset` method. It takes a `layout` (or type) and a asset `name`. It then stores this information in the queue. The method for finding an layout or type path has been moved out of `Records::Layout` and into `AssetBuilder` as this is where it is used.

Note: AssetBuilder will store the `type` of an asset in the struct as well. However the code to do this has not been implemented yet. Hence there is a place holder atm.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -86,3 +86,17 @@ RSpec/MessageSpies:
 # doesn't isn't that much clearer IMO.
 RSpec/NamedSubject:
   Enabled: false
+
+# Allow RSpec to suppress Output using a global variable. It already used a global
+# variable previously and it will not be used in the installed version of
+# Metalware. It is however useful to turn the Output off in the specs
+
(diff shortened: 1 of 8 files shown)


WilliamMcCumstie:
If you just want to review this PR @atoghill, I have rebased on the `type_from_path` work and have updated `push_asset` to unpack the type.

WilliamMcCumstie:
Ignore the code climate issues. I will fix them in the next PR.

atoghill:
Changes seem fine

PR #415, opened 2018-05-02
Add a layout from a type or a layout

atoghill:
With this PR it is now possible to add a `layout` from either a `type` or another `layout`

### DIFF
--- a/src/asset_builder.rb
+++ b/src/asset_builder.rb
@@ -42,17 +42,7 @@ module Metalware
     private
 
     def source_file_details(layout_or_type)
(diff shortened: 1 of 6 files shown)


src/asset_builder.rb
```diff
@@ -42,17 +42,7 @@ def edit_asset(name)
     private
 
     def source_file_details(layout_or_type)
```
WilliamMcCumstie: Now that the code has been moved to `Records::Layout`. We no longer need this method.

atoghill: Removed the method, it now calls `Records::Layout.type_or_layout` directly from `push_asset`

Issue #285, opened 2018-01-11
metal view error if no arguments provided


ColonelPanicks:
If `metal view` is run without arguments it dumps out an ambiguous error, would be more useful to show the help page
```
[root@controller repo]# metal view
error: The alces command input can only contain upper/ lower case letters and numbers. It may contain spaces and periods as delimitors.. Use --trace to view backtrace
[root@controller repo]# metal view --trace
/opt/metalware/src/command_helpers/alces_command.rb:35:in `alces_command_alpha_numeric_only': The alces command input can only contain upper/ lower case letters and numbers. It may contain spaces and periods as delimitors. (Metalware::InvalidInput)
	from /opt/metalware/src/command_helpers/alces_command.rb:19:in `alces_command'
	from /opt/metalware/src/commands/view.rb:19:in `cli_input_object'
	from /opt/metalware/src/commands/view.rb:15:in `run'
	from /opt/metalware/src/command_helpers/base_command.rb:41:in `initialize'
	from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/command.rb:217:in `new'
	from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/command.rb:217:in `call'
	from /opt/metalware/src/commander_extensions.rb:61:in `call'
	from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/command.rb:184:in `run'
	from /opt/metalware/src/commander_extensions.rb:49:in `run'
	from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/runner.rb:464:in `run_active_command'
	from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/runner.rb:66:in `run!'
	from /opt/metalware/vendor/ruby/2.4.0/bundler/gems/commander-c99203ec5a74/lib/commander/delegates.rb:15:in `run!'
	from /opt/metalware/src/cli.rb:53:in `run!'
	from /opt/metalware/src/cli.rb:48:in `run'
	from bin/metal:28:in `<main>'
```

WilliamMcCumstie:
Also it tries to convert strings to hashes for:
`nodes.assets.first.stuff.more.stuff.ip`

PR #416, opened 2018-05-02
Add a method to error if no arguments are provided

atoghill:
Currently running `metal view` with no arguments throws out the error message for an incorrect input for `[ALCES_COMMAND]`. However this isn't particularly helpful when no argument has been provided.

This PR implements a method within `View` that raises an error in this case. 

See #285 for full details on the issue


WilliamMcCumstie:
There is a simpler way to fix the issue:
https://github.com/alces-software/metalware/pull/417/commits/fc0c573acfd90028eb01dc0e7552663e19d6e99c

PR #420, opened 2018-05-03
Update Asset Types with Further Information

ColonelPanicks:


ColonelPanicks:
Force pushed an update with a compliant comment.

WilliamMcCumstie:
Thanks for the update, the only other thing would be a description for the PR. We might remember the details atm, however it is useful if we want to review it in 6 months time.

Issue #433, opened 2018-07-25
Rendering named zone files with invalid comment character


bobwhitelock:
As originally raised at https://alces.slack.com/archives/CBADGPNRK/p1532445688000567 - essentially, and as according to https://www.centos.org/docs/5/html/Deployment_Guide-en-US/s1-bind-zone.html, in `named` zone files "Comments can be placed after semicolon characters (;)". However in Metalware we currently render these with `#` as the comment character, making these files invalid and causing `named` to fail to start when you attempt to `metal sync` these rendered managed files.

bobwhitelock:
(I'm about to fix this, but logging this now so I can reference the issue).

PR #435, opened 2018-07-25
Eval command

### LINKED ISSUE #431: Couple of issues with using `instance_eval` in `metal view`
Following the merging of https://github.com/alces-software/metalware/pull/430 there have been a few issues.

From Slack:

1. Both of these used to show the full namespace for a node:

    ```ruby
    [root@controller [asset-example] ~]# metal view node.node780
    error: Error, a Node is not in scope. Use --trace to view backtrace
    [root@controller [asset-example] ~]# metal view nodes.node780
    "#<Metalware::Namespaces::Node:0x0000000156a760, name: node780>"
    ```

    You can still get the same effect as the old behaviour with `metal view nodes.node780.to_h`.
    
    (from https://alces.slack.com/archives/C5FL99R89/p1529505996000158)

2. Can no longer use `metal view` like this:

    ```ruby
    [root@controller [asset-example] metalware]# metal view assets.servers.rack1-r630-chassis-780-server1
    error: undefined method `rack1' for #<Metalware::Namespaces::AssetArray:0x00000001631ce8>. Use --trace to view backtrace                                                                                                                      
    ```

    because this server name is not tokenized as a full method by Ruby. I can work around this like this atm, since the method with that name's still defined:

    ```ruby
    [root@controller [asset-example] metalware]# metal view "assets.servers.public_send('rack1-r630-chassis-780-server1')"                                                                                                                        
    {
      "manufacturer": "Dell",
      "model": "r630-server-board",
    [ ... ]
    ```
    (from https://alces.slack.com/archives/C5FL99R89/p1529575880000307)

More discussion happened in Slack.

bobwhitelock:
This PR introduces a new `metal eval` command, with the current behaviour of `metal view`, and reverts `metal view` to the behaviour it had prior to #430. This should fix #431 by giving us distinct commands for human (`metal view`) vs programmatic (`metal eval`) inspection of the Metalware namespace, which means each of these can be specialised for its purpose rather than conflicting as described in that issue. 

Based on #433.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ rsync:
 		--exclude vendor/ \
 		--exclude .bundle/ \
 		--perms \
-		. dev@${IP}:${REMOTE_DIR}
+		. root@${IP}:${REMOTE_DIR}
 
 .PHONY: watch-rsync
 watch-rsync:
--- a/spec/command_helpers/alces_command_spec.rb
+++ b/spec/command_helpers/alces_command_spec.rb
@@ -1,0 +1,59 @@
+
+# frozen_string_literal: true
+
+require 'command_helpers/alces_command'
+require 'alces_utils'
+
+RSpec.describe Metalware::CommandHelpers::AlcesCommand do
+  include AlcesUtils
+
+  let(:domain_config) { Hash.new(key: 'I am the domain config') }
+
+  let(:node) { 'node01' }
+  let(:group) { 'group1' }
+
+  AlcesUtils.mock self, :each do
+    config(alces.domain, domain_config)
+    mock_group(group)
+    mock_node(node)
+  end
+
+  #
+  # The purpose of the mixin is to provide the alces_command method
+  # However as this is a private method, it has to use send
+  #
+  def test_command(command)
+    double('TestDouble', alces: alces, raw_alces_command: command)
+      .extend(Metalware::CommandHelpers::AlcesCommand)
+      .send(:alces_command)
+  end
+
+  it 'can return the domain config' do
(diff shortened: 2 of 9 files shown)


WilliamMcCumstie:
Rebased on develop

bobwhitelock:
@WilliamMcCumstie 

> However I don't think we should colourise the output in `metal eval` as it is meant to be machine readable. The colouriser will add a bunch of control characters that are essentially meaningless to a machine.
>
> Instead `metal view` should be colourised by default (maybe a flag to turn it off). Then `metal eval` should always return regular text.

ATM for both commands we colour the output if they're outputting to a terminal, or if the `--color-output` is passed - I think in both cases this is useful/reasonable behaviour as:

1. If the command is outputting to a terminal you are (almost certainly) using it interactively, so getting colourised JSON as output is helpful as it allows you to see at a glance the output of the command. This is still the case for `metal eval` as it's useful to try things out in a terminal until you get the results you need, even if it's ultimately intended to be used by another program.

2. If the command is not outputting to a terminal (which will almost certainly be the case when using `metal eval` programmatically), you probably don't want a load of terminal control characters mixed in with the useful output, so we should suppress the colourising by default.

3. There are still cases where it is useful to get coloured output when not outputting to a terminal, e.g. to allow paging the output like `metal view nodes.first --color-output | less -R`, so we have the `--color-output` to allow this behaviour.

bobwhitelock:
Hold off on merging this for a second if you were planning to soon, may want to make another tweak :slightly_smiling_face:

bobwhitelock:
@WilliamMcCumstie change made, merge at will again

PR #436, opened 2018-07-26
Switch NetworkAdapter asset to use new format

bobwhitelock:
As discussed earlier @ColonelPanicks:

> The interface is a property of a particular node's connection to a network, via a network adapter port, rather than a property of the port itself.
> 
> Therefore we have decided to store network adapter ports as an array rather than a hash including the interface, and the canonical source for interface names will be in each node's network configs.

A pretty trivial change but couple of minor things to consider before merging:

1. Are the `port` keys in this file relevant any more? They aren't used in any of the clusters I have access to, and I'm guessing this is the same information that we now have from the (1-based) index of the port in the `ports` array.

2. Do we need to make any corresponding repo changes following this change?

### DIFF
--- a/data/asset_types/network_adapter.yaml
+++ b/data/asset_types/network_adapter.yaml
@@ -5,16 +5,12 @@ ports:
   - network: null
     mac: null
     switch: null
-    port: null
   - network: null
     mac: null
     switch: null
-    port: null
   - network: null
     mac: null
     switch: null
-    port: null
   - network: null
     mac: null
     switch: null
-    port: null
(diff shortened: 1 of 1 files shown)


ColonelPanicks:
1. I think it's safe to assume we don't actually need to specify the port numbers and, as you suggested, just use the array index as the corresponding port on the network adapter.
2. I've had a quick look and there isn't anything in the repo that looks for network adapters or would need otherwise changing

bobwhitelock:
Thanks for confirming, made that tweak and will merge now :slightly_smiling_face:

PR #438, opened 2018-08-01
Address #437 with chomp

### LINKED ISSUE #437: Weird Control Characters in PXE Config File Names
Soooooo when the `metal build` command runs for a node it outputs the hexadecimal IP of the node but seems to be appending a strange character:
```
# ls output
[root@controller [testcluster] pxelinux.cfg]# ls
0A0A0101?  0A0A0102?  0A0A1001?  default

# Different output on tab complete
[root@controller [testcluster] pxelinux.cfg]# metal render /tmp/test node001 > /var/lib/tftpboot/pxelinux.cfg/
0A0A0101^J  0A0A0102^J  0A0A1001^J  default
```

Any idea what's going on here?? I can't read the files due to the strange control characters and PXE booting is pretty integral to metalware!

ColonelPanicks:
This fixes #437

### DIFF
--- a/src/namespaces/node.rb
+++ b/src/namespaces/node.rb
@@ -55,7 +55,7 @@ module Metalware
       end
 
       def hexadecimal_ip
-        @hexadecimal_ip ||= SystemCommand.run "gethostip -x #{name}"
+        @hexadecimal_ip ||= SystemCommand.run("gethostip -x #{name}").chomp
       end
 
       def build_method


WilliamMcCumstie:
This looks like a decent fix. However it begs the question on how we deal with the new line characters.

I'm of the opinion that `SystemCommand` is behaving as expected as most system commands do end in a newline character. Thus it is the responsibility of the calling method to chomp the character off.

However, there maybe other cases that don't do this. So we might come across a similar bug again. @bobwhitelock, you have an opinion?

bobwhitelock:
This looks fine to me for now, but open to being convinced that we should do something else if there's a reason to :slightly_smiling_face:

WilliamMcCumstie:
I'll merge this through now. We can revisit the issue if it crops up again

PR #440, opened 2018-08-08
Fix build methods

### LINKED ISSUE #439: Basic build_method not rendering appropriate build files
Original issue for adding build_method - https://github.com/alces-software/metalware/issues/145

When using basic build_method on a node I'd expect it to render `basic/default` from the metalware repo however what seems to happen is that it just runs a pxe_kickstart build method (as I can see it putting in place `rendered/kickstart/nodename` and `/var/lib/tftpboot/pxelinux.cfg/nodename`). 

Am I misunderstanding the basic build method implementation? The idea was that it could create a file we could curl to begin the installation.

bobwhitelock:
This PR contains a fix for #439, which is also a fix for the the more general issue that all non-Pxelinux/Kickstart build methods were broken. Also included are some `setup` script tweaks to make this possible to work the first time it is run after syncing Metalware to a machine.

@ColonelPanicks can you try this out and let me know if this does/doesn't fix your original issue? @WilliamMcCumstie if this does fix #439 then I'll merge it unless there's some underlying reason why we thought we didn't need the original `to_sym`ing (and so this might need deeper investigation; see https://github.com/alces-software/metalware/commit/e01b30ef39ff9df3ef9c8032fba8e8e685757068#r30002376).

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -27,7 +27,14 @@ IFS=$'\n\t'
 # Install RVM.
 gpg2 --keyserver hkp://pgp.mit.edu --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
 curl -sSL https://get.rvm.io | bash -s stable
-# XXX Need to log out and in again here for script to work on first run.
+
+# Enable RVM, which is less strict about letting commands fail, variables be
+# unset etc. than this script.
+set +euo pipefail
+# shellcheck disable=SC1091
+source /etc/profile.d/rvm.sh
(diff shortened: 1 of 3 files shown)


bobwhitelock:
@ColonelPanicks does this fix your issue? :slightly_smiling_face:

ColonelPanicks:
@bobwhitelock I let this slip, will test it out today and let you know.

bobwhitelock:
Thanks both

PR #441, opened 2018-08-13
Update changelog for 2018.2.0-rc3

bobwhitelock:
Plus minor related changes; see commits.


bobwhitelock:
Noticed I've used the wrong version name and things are missing from the changelog as it wasn't updated for the last stable release or recent rcs; going to revert this and fix.

PR #447, opened 2018-08-16
Allow --answers for local

ColonelPanicks:
Fixes #447


bobwhitelock:
Nice, can be merged once CI passes (which may require restarting it) :+1:.

ColonelPanicks:
Brill. Would it be possible to put this into a 2018.3.1 release?

bobwhitelock:
@ColonelPanicks sure :slightly_smiling_face: - do you want to do that, should be fairly straightforward as described in https://github.com/alces-software/metalware/blob/develop/docs/release-process.md.

PR #450, opened 2018-10-05
Find and remove unused code in Metalware

bobwhitelock:
Use https://github.com/joshuaclayton/unused to find and then remove various unused code in Metalware.

This seems worth doing generally as no point keeping around unused cruft in the codebase, but I'm specifically doing this now to get things in a clean state so I can then more easily use the same tool to incrementally identify further unused code while splitting many parts of Metalware out into (codename) Underware.

Going to merge this straight through when CI passes since it's just some tidying up, but feel free to let me know if you have any opinions.


bobwhitelock:
Merging this through now, the only failure looks  spurious and unrelated.

PR #456, opened 2018-10-29
Adapt Metalware to use Underware

bobwhitelock:
Many, many changes made to adapt Metalware to use [Underware](https://github.com/alces-software/underware). Many things have been removed as they now live in Underware, various other things have been adapted to reduce duplication and coupling with Underware. Also did lots of tidying at the same time, since I was taking a look over the whole codebase while doing this. See the many commits for the many details.


bobwhitelock:
Merging through now as CI has already passed, it's just picked up on my tiny amendment to a commit and is running again.

